PR #62, opened 2022-10-13
Testnet deployment

kkirka:
This PR adds BSC testnet deployment addresses and ABIs generaated with hardhat-deploy. It also fixes a small bug in one of the deployment scripts (see the first commit for details)


coreyar:
> Re: creating `networks/` folder – I strongly object storing the same data in different files in the repo, because at some point the addresses in `deployments/` folder will differ from the addresses in `networks/` folder. This is how people work – someone will forget to update something. If we want to include `networks/` in a package, why can't we autogenerate `networks/` from `deployments/`?

yes I believe that was what Simeone was working on with JQ

0xlucian:
@coreyar  Ahmm not really.. I saw JQ from @kkirka and was playing with it... but we can do this automation for sure

0xlucian (approved):

contracts/test/VBep20Harness.sol
```diff
@@ -7,7 +7,7 @@ import "../VBep20Delegate.sol";
 import "../Governance/AccessControlManager.sol";
 import "./ComptrollerScenario.sol";
 
-contract VBep20Harness is VBep20Immutable {
+contract VBEP20Harness is VBep20Immutable {
```
kkirka, line 10: Why though?

coreyar: It resolves a typescript compile error originating from generated typechain that complains filename casing doesn't match imports

PR #66, opened 2022-10-22
Emit poolid with name set event

coreyar:
## Description

Events related to the pool should emit the same identifier for consistency. This id needs to match the id in the subgraph for easy querying. 

Since other events emit the pool index it is logic to use this as the id and emit it with all pool events. This PR updates the pool name set event to use the index as the identifier

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,8 @@ RUN wget https://github.com/ethereum/solidity/releases/download/v0.8.13/solc-sta
 RUN mkdir -p /usr/app
 WORKDIR /usr/app
 
+ENV NODE_OPTIONS=--max-old-space-size=4096 
+
 # First add deps
 RUN npm install -g yarn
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN wget https://github.com/ethereum/solidity/releases/download/v0.8.13/solc-sta
 RUN mkdir -p /usr/app
 WORKDIR /usr/app
 
+ENV NODE_OPTIONS=--max-old-space-size=4096
```
0xlucian, line 15: Why do we increase heap limit here? Is this error caused by the changes of the Event argument type?

coreyar: This error is not related to changes in this PR and affects [other PRs](https://github.com/VenusProtocol/isolated-pools/pull/64) as well.

I am not clear on the original of the error but it is caused by javascript running out of allocated memory when compiling the contracts with hardhat

0xlucian (approved):

PR #73, opened 2022-10-28
refactor: make pool idx internal, comptroller as id

coreyar:
## Description

Events and functions accessing pools through index
have been refactored to use comptroller address

Accessers using poolID are removed
Mapping on shortfall for indexing with poolId has been removed
PoolId is no longer returned as it is used internally for looping over all pools only

### DIFF
--- a/contracts/Pool/PoolRegistry.sol
+++ b/contracts/Pool/PoolRegistry.sol
@@ -93,12 +93,13 @@ contract PoolRegistry is OwnableUpgradeable {
     /**
      * @dev Maps venus pool id to metadata
      */
-    mapping(uint256 => VenusPoolMetaData) public metadata;
+    mapping(address => VenusPoolMetaData) public metadata;
 
     /**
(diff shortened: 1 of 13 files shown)


contracts/Pool/PoolRegistry.sol
```diff
@@ -108,7 +109,7 @@ contract PoolRegistry is OwnableUpgradeable {
     /**
     * @dev Maps comptroller address to Venus pool Index.
     */
-    mapping(address => uint256) private _poolByComptroller;
+    mapping(address => VenusPool) private _poolByComptroller;
```
defistar, line 112: @coreyar current version of code, had this storage mapping to lookup `poolId` from `comptrollerAddress`
in existing code its mapping for poolId -> VenusPoolStruct
should we retain same way ?


now that this PR is to use `poolId` as primary reference while looking-up the Pool
and mapping has comptrollerAddress -> VenusPool Struct

defistar (approved):

PR #80, opened 2022-11-02
 correctly ignore DS_Store

coreyar:
## Description

Correctly ignore DS_Store


coreyar:
> Shouldn't this be in the [global .gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) rather than in the repo?

That is interesting. Let's take this for now we can discuss and establish a standard if we want to set the expectation that team members include this in a local global git config

PR #83, opened 2022-11-04
[VEN-727] Replace Unitroller with an OpenZeppelin proxy

kkirka:
This PR replaces a custom proxy for the comptroller with a standard openzeppelin implementation. The latter is widely used and better supported by the tooling.

Apart from this, this PR has some style fixes for the contracts.


contracts/Comptroller.sol
```diff
@@ -432,6 +439,11 @@ contract Comptroller is
         if (redeemTokens == 0 && redeemAmount > 0) {
             revert("redeemTokens zero");
         }
+
+        // Shh - we don't ever want this hook to be marked pure
```
stubbi, line 443: should these lines be here then?

0xlucian (approved):

PR #91, opened 2022-11-17
ci: run tests in isolation

coreyar:
## Description

Run tests in CI in isolation so they won't fail due to external dependency failures


.github/workflows/ci.yaml
```diff
@@ -46,15 +44,9 @@ jobs:
...
-      - name: "Create env file"
-        run: |
-          touch .env
-          echo FORK_MAINNET=${{ secrets.FORK_MAINNET }} >> .env
-          echo QUICK_NODE_KEY=${{ secrets.QUICK_NODE_KEY }} >> .env
-
       - name: Run hardhat compile and tests coverage
         run: |
-          source .env
+          source .env.example
```
0xlucian, line 49: shouldn't we source .env.test? as it will be commited in the repo

tests/hardhat/Fork/RiskFund.ts
```diff
@@ -338,13 +352,19 @@ const riskFundFixture = async (): Promise<void> => {
   await comptroller2Proxy._setPriceOracle(priceOracle.address);
   await comptroller3Proxy._setPriceOracle(priceOracle.address);
 
-  pancakeSwapRouter = await PancakeRouter__factory.connect("0x10ED43C718714eb63d5aA57B78B54704E256024E", admin);
+  if (FORK_MAINNET) {
+    pancakeSwapRouter = await PancakeRouter__factory.connect("0x10ED43C718714eb63d5aA57B78B54704E256024E", admin);
+  } else {
+    const pancakeSwapRouterFactory = await smock.mock<PancakeRouter__factory>("PancakeRouter");
+    pancakeSwapRouter = await pancakeSwapRouterFactory.deploy("0x10ED43C718714eb63d5aA57B78B54704E256024E", admin.address);
+    await pancakeSwapRouter.deployed();
```
0xlucian, line 360:
```suggestion
   pancakeSwapRouter = await smock.fake<PancakeRouter>("PancakeRouter");
    pancakeSwapRouter.swapExactTokensForTokens
      .returns((amountIn: any, amountOutMin: any, path: any, to: any, deadline: any) => [0, amountIn] );
```

0xlucian (changes requested):
The unit tests were failing. Please look at the suggested changes and after applying try to run tests.

.env.example
```diff
@@ -1,5 +1,3 @@
 ETHERSCAN_API_KEY=ABC123ABC123ABC123ABC123ABC123ABC1
-ROPSTEN_URL=https://eth-ropsten.alchemyapi.io/v2/<YOUR ALCHEMY KEY>
 PRIVATE_KEY=0xabc123abc123abc123abc123abc123abc123abc123abc123abc123abc123abc1
-FORK_MAINNET=
-QUICK_NODE_KEY=
```
defcon022, line 5: I think we should keep the variables in the env example file. If we dont want to run fork tests at a momwnt simply we can pass `FORK_MAINNET` as false in env to do that.

.github/workflows/ci.yaml
```diff
@@ -46,15 +44,9 @@ jobs:
       - name: Install solidity
         run: yarn global add solc@0.8.13
 
-      - name: "Create env file"
-        run: |
-          touch .env
-          echo FORK_MAINNET=${{ secrets.FORK_MAINNET }} >> .env
-          echo QUICK_NODE_KEY=${{ secrets.QUICK_NODE_KEY }} >> .env
```
defcon022, line 53: Here also we should keep the set of the env in theyam file and just pass `FORK_MAINNET` as false in github secrets to avoid running mainnet tests.

tests/hardhat/Fork/RiskFund.ts
```diff
@@ -44,13 +44,16 @@ let whitePaperRateFactory: WhitePaperInterestRateModelFactory;
 let fakeAccessControlManager: FakeContract<AccessControlManager>;
 let protocolShareReserve: ProtocolShareReserve;
 let riskFund: RiskFund;
-let pancakeSwapRouter: PancakeRouter;
+let pancakeSwapRouter: PancakeRouter | MockContract<PancakeRouter>;
```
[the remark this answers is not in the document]
coreyar: I think the Mock contract is preferable in general because it allows for real contract code to be used by default rather than mocking the entire contract

.github/workflows/ci.yaml
coreyar: We don't need this anymore because we have a test env file setup that is being sourced in the following step

0xlucian (approved):

defcon022 (approved):

PR #95, opened 2022-11-29
Audit improvements, low-hanging fruits

kkirka:
This PR implements the following improvement suggestions from the internal audit report:

- **VENUS-IMP-001** ABICoder V2 Enabled By The Default After 0.8.x
- **VENUS-IMP-003** `++i` is cheaper than `i++`
- **VENUS-IMP-004** Caching array length can save gas
- **VENUS-IMP-005** Events are not indexed
- **VENUS-IMP-006** No need to initialize variables with default value
- **VENUS-IMP-008** Dont compare boolean expressions to boolean literals
- **VENUS-IMP-010** Use of power operations instead of their exponential form

Other improvement suggestions either need further discussion or will be addressed in the scope of a bigger refactoring (e.g. moving from error strings to custom error messages).

Note that "caching the error length" commit also moves some logic to separate internal functions to prevent stack too deep errors, so these are not purely stylistic changes and should be reviewed with full attention.

### DIFF
--- a/contracts/Comptroller.sol
+++ b/contracts/Comptroller.sol
@@ -133,7 +133,7 @@ contract Comptroller is
         uint256 len = vTokens.length;
 
         uint256[] memory results = new uint256[](len);
-        for (uint256 i = 0; i < len; ++i) {
+        for (uint256 i; i < len; ++i) {
             VToken vToken = VToken(vTokens[i]);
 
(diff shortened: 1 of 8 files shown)


coreyar (approved):

0xlucian (approved):
lgtm

PR #98, opened 2022-11-30
Set unlimited borrow cap as maxint

coreyar:
## Description

Updates borrow cap so 0 means borrowing is not allowed and maxint256 is unlimited borrowing

### DIFF
--- a/deploy/003-deploy-pools.ts
+++ b/deploy/003-deploy-pools.ts
@@ -121,6 +121,11 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     tokenImplementation_: tokenImplementation.address,
   });
   await tx.wait();
+
+  comptroller1Proxy._setMarketBorrowCaps(
+    [tokenImplementation.address],
+    ["0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"],
+  );
 };
 
 func.tags = ["Pools"];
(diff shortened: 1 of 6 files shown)


deploy/003-deploy-pools.ts
```diff
@@ -121,6 +121,11 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     tokenImplementation_: tokenImplementation.address,
   });
   await tx.wait();
+
+  comptroller1Proxy._setMarketBorrowCaps(
+    [tokenImplementation.address],
+    ["0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"],
+  );
```
kkirka, line 128: Shouldn't we set supply caps as well somewhere?

coreyar: The supply cap isn't using 0 for infinite supplies. The supply cap uses 0 as minting not allowed.

The borrow cap works differently because it skips code if the cap is infinite. Whereas supply always checks the cap

kkirka: Yeah, I mean this line allows to borrow tokens, but I can't find a line that sets supply cap from zero to 0xfff...fff. This means the supply should be impossible, because, as you said, 0 means minting not allowed 🤔

kkirka (approved):

PR #99, opened 2022-11-30
Add reward speed view

coreyar:
## Description

Adds a view function to return the total reward speed for a market


contracts/Comptroller.sol
```diff
@@ -1125,6 +1140,25 @@ contract Comptroller is
         return (uint256(Error.NO_ERROR), seizeTokens);
     }
 
+    /**
+     * @notice Returns XVS reward speed given a vToken
+     * @param vToken The vToken to get the reward speeds for
+     * @return rewardSpeeds Array of total supply and borrow speeds and reward token for all reward distributors
+     */
+    function getXVSRewardsByMarket(address vToken) external view returns (RewardSpeeds[] memory rewardSpeeds) {
```
kkirka, line 1148: These wouldn't necessarily be XVS, so I propose we just drop the token name here

```suggestion
    /**
     * @notice Returns reward speed given a vToken
     * @param vToken The vToken to get the reward speeds for
     * @return rewardSpeeds Array of total supply and borrow speeds and reward token for all reward distributors
     */
    function getRewardsByMarket(address vToken) external view returns (RewardSpeeds[] memory rewardSpeeds) {
```

kkirka (approved):

PR #111, opened 2022-12-05
fix!: [012] add _disableInitializers to all upgradeable contracts

kkirka:
Problem: Uninitialized implementation may be initialized by an attacker if the contract is not deployed in the same transaction as proxy deployment or upgrade.

Solution: Add _disableInitializers call to the constructors. This affects tests, so this commit also updates tests (with some refactoring to replace unnecessary deployments with fakes).

Note that these changes are **BREAKING.** The corresponding commits are marked with `!`

Fixes the internal audit finding **012**.

**Note:** The PR is based on https://github.com/VenusProtocol/isolated-pools/pull/110 and marked as draft. The base should be changed to main and the PR can be marked as ready after https://github.com/VenusProtocol/isolated-pools/pull/110 is merged.

### DIFF
--- a/tests/hardhat/Comptroller/accountLiquidityTest.ts
+++ b/tests/hardhat/Comptroller/accountLiquidityTest.ts
@@ -2,7 +2,7 @@ import { FakeContract, MockContract, smock } from "@defi-wonderland/smock";
 import { loadFixture, setBalance } from "@nomicfoundation/hardhat-network-helpers";
 import chai from "chai";
 import { BigNumber, BigNumberish, Signer } from "ethers";
-import { ethers } from "hardhat";
+import { ethers, upgrades } from "hardhat";
 
 import { convertToUnit } from "../../../helpers/utils";
 import {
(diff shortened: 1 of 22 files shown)


tests/hardhat/Comptroller/accountLiquidityTest.ts
```diff
@@ -28,9 +28,11 @@ interface AccountLiquidityTestFixture {
 async function makeComptroller(): Promise<AccountLiquidityTestFixture> {
   const poolRegistry = await smock.fake<PoolRegistry>("PoolRegistry");
   const accessControl = await smock.fake<AccessControlManager>("AccessControlManager");
-  const comptrollerFactory = await smock.mock<Comptroller__factory>("Comptroller");
-  const comptroller = await comptrollerFactory.deploy(poolRegistry.address, accessControl.address);
-  await comptroller.initialize();
+  const Comptroller = await smock.mock<Comptroller__factory>("Comptroller");
```
coreyar, line 31: I haven't pushed my changes but essentially I broke this function and others out into its own file so that we don't have it duplicated in ever test file

0xlucian (approved):

PR #112, opened 2022-12-05
fix: [017] approve zero before approving nonzero amount

kkirka:
Problem: Non-standard tokens like USDT will revert the transaction when a contract tries to approve an allowance when the spender allowance is already set a non-zero value. This can potentially lead to denial of service.

Solution: Approve zero before setting allowance to nonzero value.


defcon022:
can we have a fork test for `USDT` token on mainnet BSC chain to see if `swapPoolsAssets` is working fine.

kkirka:
> can we have a fork test for `USDT` token on mainnet BSC chain to see if `swapPoolsAssets` is working fine?

@defcon022 I don't think I have capacity for this rn, but I agree having such a test would be nice. Can you or someone else from the team contribute?

coreyar (approved):

coreyar (approved):

PR #125, opened 2022-12-14
refactor: refactor hardfork setup to take a URL instead of API key

coreyar:
## Description

Updates fork test setup to use its own network and take an RPC url instead of secret key


hardhat.config.ts
```diff
@@ -150,7 +150,21 @@ const config: HardhatUserConfig = {
...
+      loggingEnabled: false,
+      live: false,
+    },
+    mainnetfork: {
+      allowUnlimitedContractSize: false,
+      loggingEnabled: false,
+      url: process.env.FORK_MAINNET_RPC as string,
+      forking: {
+        url: process.env.FORK_MAINNET_RPC as string,
+        blockNumber: 21068448,
```
kkirka, line 164: I guess we could use an env var here as well?

It's really sad hardhat config can't accept parameters based on the test we're running. It would be nice if we could specify the fork block right in the test suite – but afaik hh doesn't support this because it doesn't expose the network setup to tests.

tests/hardhat/Fork/RiskFund.ts
```diff
@@ -136,7 +136,7 @@ const riskFundFixture = async (): Promise<void> => {
   shortfall = await smock.fake<Shortfall>("Shortfall");
   await admin.sendTransaction({
     to: shortfall.address,
-    value: ethers.utils.parseEther("1"), // 1 ether
+    value: ethers.utils.parseEther(".5"), // 1 ether
```
kkirka, line 139: Now the comment is no longer true 🤔

kkirka (approved):
Left some comments but overall LGTM

hardhat.config.ts
```diff
@@ -150,7 +150,21 @@ const config: HardhatUserConfig = {
     ],
   },
   networks: {
-    hardhat: isFork(),
+    hardhat: {
+      allowUnlimitedContractSize: true,
+      loggingEnabled: false,
+      live: false,
+    },
+    mainnetfork: {
```
defcon022, line 158: Is adding a network (`mainnetfork`) to run forked tests working here?
Because I tried the same thing as @coreyar mentioned in the [PR#181 comments](https://github.com/VenusProtocol/venus-protocol/pull/181#discussion_r1048930239) for Venus Protocol, and it was not working for me there.

coreyar: yes it works

hardhat.config.ts
coreyar: sounds good

PR #143, opened 2023-01-24
[VEN-890] Multiple user engagement integration tests

UrAvgDeveloper:
This PR consists of the combined tickets [VEN-886], [VEN-887], [VEN-888], [VEN-889], and [VEN-890], where [VEN-890] is separated and includes a test case that performs activities in a single block, [VEN-888] and [VEN-889] do not have separate test cases because these are included in all of the cases, and the remaining tickets are in the pending stage due to the deployment script refactoring.

## Checklist


tests/integration/index.ts
```diff
@@ -6,22 +6,43 @@ import { ethers } from "hardhat";
...
 
 import { convertToUnit, scaleDownBy } from "../../helpers/utils";
-import { AccessControlManager, Comptroller, MockToken, PoolRegistry, PriceOracle, VToken } from "../../typechain";
+import {
+  AccessControlManager,
+  Comptroller,
+  MockToken,
+  PoolRegistry,
+  PriceOracle, // RiskFund,
+  // Shortfall,
```
0xlucian, line 15:
```suggestion
```

0xlucian (changes requested):
Should we merge this before the exchange rate fix we are waiting? I can see some commented out code waiting for this fix?

tests/integration/index.ts
chechu: I would wait until having the fix

PR #144, opened 2023-01-24
feat: Mint and Redeem events emit the updated account balance

gleiser-oliveira:
## Description

This PR aims to make the Mint and Redeem VToken events also return the updated value for the account balance. This can make the subgraph save calling the `balanceOf` function everytime it indexes these events.

### DIFF
--- a/contracts/VToken.sol
+++ b/contracts/VToken.sol
@@ -552,7 +552,7 @@ contract VToken is Ownable2StepUpgradeable, VTokenInterface, ExponentialNoError,
         accountTokens[minter] = accountTokens[minter] + mintTokens;
 
         /* We emit a Mint event, and a Transfer event */
-        emit Mint(minter, actualMintAmount, mintTokens);
+        emit Mint(minter, actualMintAmount, mintTokens, accountTokens[minter]);
         emit Transfer(address(0), minter, mintTokens);
     }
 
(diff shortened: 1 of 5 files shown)


tests/hardhat/Tokens/mintAndRedeemTest.ts
```diff
@@ -231,7 +231,7 @@ describe("VToken", function () {
 
       expect(result) // eslint-disable-line @typescript-eslint/no-floating-promises
```
chechu, line 232: We need `await` here. Otherwise no errors will be triggered

tests/hardhat/Tokens/mintAndRedeemTest.ts
```diff
@@ -231,7 +231,7 @@ describe("VToken", function () {
...
         .to.emit(vToken, "Mint")
-        .withArgs(minterAddress, mintAmount, mintTokens);
+        .withArgs(minterAddress, mintAmount, mintTokens, mintAmount);
```
chechu, line 234: Assuming the initial balance of vTokens for the user was 0, the last argument of the event will be `minTokens` (the updated balance of vTokens)

chechu (approved):

contracts/VToken.sol
```diff
@@ -662,7 +662,7 @@ contract VToken is Ownable2StepUpgradeable, VTokenInterface, ExponentialNoError,
 
         /* We emit a Transfer event, and a Redeem event */
         emit Transfer(redeemer, address(this), redeemTokens);
-        emit Redeem(redeemer, redeemAmount, redeemTokens);
+        emit Redeem(redeemer, redeemAmount, redeemTokens, accountTokens[redeemer]);
```
0xlucian, line 665: Reading from storage uses alot of gas. Would be more gas efficient if we do something like:
```
uint256 balanceAfter = accountTokens[redeemer] - redeemTokens;
accountTokens[redeemer] = balanceAfter;
emit Redeem(redeemer, redeemAmount, redeemTokens, balanceAfter);
```

0xlucian (changes requested):

0xlucian (approved):

chechu (approved):

narayanprusty (approved):

PR #149, opened 2023-01-27
[VEN-936] - Fixed exchange rate decimals issue.

defcon022:
## Description

This PR fixed exchange rate decimals issue. Attached ticket: VEN-936

## Checklist

### DIFF
--- a/contracts/Pool/PoolRegistry.sol
+++ b/contracts/Pool/PoolRegistry.sol
@@ -249,7 +249,7 @@ contract PoolRegistry is Ownable2StepUpgradeable {
             input.asset,
             comptroller,
             rate,
-            10**input.decimals,
+            10**18,
             input.name,
             input.symbol,
             input.decimals,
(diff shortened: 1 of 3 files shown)


contracts/Pool/PoolRegistry.sol
```diff
@@ -249,7 +249,7 @@ contract PoolRegistry is Ownable2StepUpgradeable {
             input.asset,
             comptroller,
             rate,
-            10**input.decimals,
+            10**18,
```
0xlucian, line 252: VTokens should have 8 decimals. Shouldn't we just set the exchange rate in a way that when we deposit a token lets say with 18 decimals to result in 1e8 vTokens?

0xlucian: In other words, exchange rate should be computed out of the decimal places of the underlying token... We should not assume that every token will have 18 decimal places right? 
We only assume that vTokens in isolated-pools will always have 8 decimals, but this is not the case for the underlying imo

0xlucian (approved):

0xlucian (approved):

0xlucian:
CLosing this PR since the changes has been implemented in `refactor/deployments` branch

PR #152, opened 2023-01-31
Increase Rewards Tests Coverage

narayanprusty:
## Description

Resolves #VEN-825

## Checklist

- [x] If I added new functionality, I added tests covering it.
- [x] If I fixed a bug, I added a test preventing this bug from silently reappearing again.
- [x] My contribution follows [Venus contribution guidelines](docs/CONTRIBUTING.md).


defcon022 (approved):

chechu (commented):
Would it make sense to check the specific values we would expect? Not only checking they are bigger than 0. These values shouldn't change with changes in the implementation. They are defined by business rules. This way we would check that we implemented these business rules properly

tests/hardhat/Rewards.ts
```diff
@@ -39,11 +40,10 @@ let fakePriceOracle: FakeContract<PriceOracle>;
 let fakeAccessControlManager: FakeContract<AccessControlManager>;
 
 describe("Rewards: Tests", async function () {
-  const RewardsDistributor = await ethers.getContractFactory("RewardsDistributor");
   /**
    * Deploying required contracts along with the poolRegistry.
    */
-  before(async function () {
+  beforeEach(async function () {
```
kkirka, line 46: Might be quite slow. Maybe move it into a fixture?

narayanprusty:
> Would it make sense to check the specific values we would expect? Not only checking they are bigger than 0. These values shouldn't change with changes in the implementation. They are defined by business rules. This way we would check that we implemented these business rules properly

I agree. I have calculated the rewards and updated the tests.

tests/hardhat/Rewards.ts
narayanprusty: Done

narayanprusty: Done

chechu (approved):

PR #159, opened 2023-02-06
Hacken audit fix

UrAvgDeveloper:
## Description
VEN-1038: L06. Division by Zero
VEN-1039: L07. Style Guide Violation
VEN-1040: L08. Functions that Can Be Declared External
VEN-1041: L09. Boolean Equality
VEN-1042: L10. Variable Shadowing
VEN-1043: L11. Unindexed Events
VEN-1046: L14. Use of Hard-Coded Values

## Checklist

- [x] If I added new functionality, I added tests covering it.
- [x] If I fixed a bug, I added a test preventing this bug from silently reappearing again.
- [x] My contribution follows [Venus contribution guidelines](docs/CONTRIBUTING.md).

### DIFF
--- a/contracts/BaseJumpRateModelV2.sol
+++ b/contracts/BaseJumpRateModelV2.sol
@@ -87,6 +87,26 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
     }
 
     /**
+     * @notice Calculates the current supply rate per block
+     * @param cash The amount of cash in the market
+     * @param borrows The amount of borrows in the market
+     * @param reserves The amount of reserves in the market
+     * @param reserveFactorMantissa The current reserve factor for the market
+     * @return The supply rate percentage per block as a mantissa (scaled by BASE)
+     */
+    function getSupplyRate(
+        uint256 cash,
+        uint256 borrows,
+        uint256 reserves,
+        uint256 reserveFactorMantissa
+    ) public view virtual override returns (uint256) {
+        uint256 oneMinusReserveFactor = BASE - reserveFactorMantissa;
+        uint256 borrowRate = getBorrowRateInternal(cash, borrows, reserves);
+        uint256 rateToPool = (borrowRate * oneMinusReserveFactor) / BASE;
+        return (utilizationRate(cash, borrows, reserves) * rateToPool) / BASE;
+    }
+
+    /**
      * @notice Calculates the utilization rate of the market: `borrows / (cash + borrows - reserves)`
      * @param cash The amount of cash in the market
      * @param borrows The amount of borrows in the market
@@ -132,26 +152,6 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
             }
             return ((excessUtil * jumpMultiplierPerBlock) / BASE) + normalRate;
         }
-    }
-
-    /**
-     * @notice Calculates the current supply rate per block
-     * @param cash The amount of cash in the market
-     * @param borrows The amount of borrows in the market
-     * @param reserves The amount of reserves in the market
-     * @param reserveFactorMantissa The current reserve factor for the market
-     * @return The supply rate percentage per block as a mantissa (scaled by BASE)
-     */
-    function getSupplyRate(
-        uint256 cash,
-        uint256 borrows,
-        uint256 reserves,
-        uint256 reserveFactorMantissa
-    ) public view virtual override returns (uint256) {
-        uint256 oneMinusReserveFactor = BASE - reserveFactorMantissa;
-        uint256 borrowRate = getBorrowRateInternal(cash, borrows, reserves);
-        uint256 rateToPool = (borrowRate * oneMinusReserveFactor) / BASE;
-        return (utilizationRate(cash, borrows, reserves) * rateToPool) / BASE;
     }
 
     /**
--- a/contracts/BaseJumpRateModelV2.sol
+++ b/contracts/BaseJumpRateModelV2.sol
@@ -123,8 +123,6 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
             return 0;
         }
 
-        // At this point, borrows should always be greater than zero; the condition below ensures that (cash + borrows - reserves) should never be zero.
-        require(cash >= reserves, "BaseJumpRateModelV2 : division by zero error");
         return (borrows * BASE) / (cash + borrows - reserves);
     }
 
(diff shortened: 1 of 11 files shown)


contracts/BaseJumpRateModelV2.sol
```diff
@@ -103,6 +103,8 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
             return 0;
         }
 
+        // At this point, borrows should always be greater than zero; the condition below ensures that (cash + borrows - reserves) should never be zero.
```
chechu, line 126: I think this constraint is harder than we need and it could block almost any operation in the market. For example, if we don't reduce the reserve for a long time, or if the liquidation is low because we have a big demand of borrows, we'll have `cash < reserves` and therefore we won't accept new supplies.

Maybe `require(cash + borrows - reserves > 0, "BaseJumpRateModelV2 : division by zero error");` could be more precise, avoiding the division by zero but at the same time allowing supplies in periods with low liquidation.

Moreover, I would add the same require statement in WhitePaperInterestRateModel.sol

What do you think?

chechu:
For the style guides, I would update the `.solhint.json` file, setting this values:
```
{
  "extends": "solhint:recommended",
  "plugins": ["prettier"],
  "rules": {
    "code-complexity": ["error", 8],
    "compiler-version": ["error", ">=0.5.0"],
    "const-name-snakecase": "warn",
    "constructor-syntax": "error",
    "func-visibility": ["error", { "ignoreConstructors": true }],
    "max-line-length": ["error", 175],
    "not-rely-on-time": "warn",
    "reason-string": ["warn", { "maxLength": 64 }],
    "ordering": "error"
  }
}
```

We will need to add the dependency `solhint-plugin-prettier` too. After setting this file, the command `npm run lint:sol` will show how there are more cases where we have to move the functions because the order is wrong.

contracts/VTokenInterfaces.sol
```diff
@@ -157,30 +157,45 @@ abstract contract VTokenInterface is VTokenStorage {
     /**
      * @notice Event emitted when interest is accrued
      */
-    event AccrueInterest(uint256 cashPrior, uint256 interestAccumulated, uint256 borrowIndex, uint256 totalBorrows);
+    event AccrueInterest(
+        uint256 indexed cashPrior,
+        uint256 indexed interestAccumulated,
+        uint256 indexed borrowIndex,
+        uint256 totalBorrows
+    );
```
chechu, line 165: I don't see a common use case for indexing `uint256`. Is someone going to search `AccrueInterest` events where the `cashPrior` is a specific amount? Considering the extra gas cost of indexed parameters, and the high frequency of this event, I think I wouldn't index any parameter in this event.

contracts/VTokenInterfaces.sol
```diff
@@ -157,30 +157,45 @@ abstract contract VTokenInterface is VTokenStorage {
...
 
     /**
      * @notice Event emitted when tokens are minted
      */
-    event Mint(address minter, uint256 mintAmount, uint256 mintTokens, uint256 accountBalance);
+    event Mint(address indexed minter, uint256 mintAmount, uint256 indexed mintTokens, uint256 indexed accountBalance);
```
chechu, line 170: I wouldn't index uint parameters. This comment is applicable for the rest of events.

contracts/BaseJumpRateModelV2.sol
UrAvgDeveloper: Right, but can we skip these changes because we're using a solidity version higher than 0.8 and this can be handled by default?

UrAvgDeveloper:
> For the style guides, I would update the `.solhint.json` file, setting this values:
> 
> ```
> {
>   "extends": "solhint:recommended",
>   "plugins": ["prettier"],
>   "rules": {
>     "code-complexity": ["error", 8],
>     "compiler-version": ["error", ">=0.5.0"],
>     "const-name-snakecase": "warn",
>     "constructor-syntax": "error",
>     "func-visibility": ["error", { "ignoreConstructors": true }],
>     "max-line-length": ["error", 175],
>     "not-rely-on-time": "warn",
>     "reason-string": ["warn", { "maxLength": 64 }],
>     "ordering": "error"
>   }
> }
> ```
> 
> We will need to add the dependency `solhint-plugin-prettier` too. After setting this file, the command `npm run lint:sol` will show how there are more cases where we have to move the functions because the order is wrong.

If we keep a strict check for function ordering, we must change the order of the __gap[x] variable, which is openZepplin standard to avoid storage collisions and to place this variable at the end of the contract.

chechu:
> If we keep a strict check for function ordering, we must change the order of the __gap[x] variable, which is openZepplin standard to avoid storage collisions and to place this variable at the end of the contract.

I thought we had to put the `__gap` variable at the end of the *state variables*, not at the end of the contract. I think there is not collisions between the state variables and the rest of the definitions

contracts/RiskFund/ProtocolShareReserve.sol
```diff
@@ -14,6 +14,8 @@ contract ProtocolShareReserve is Ownable2StepUpgradeable, ExponentialNoError, Re
 
     address private protocolIncome;
     address private riskFund;
+    uint256 private constant protocolSharePercentage = 70;
```
chechu, line 17:
```suggestion
    // Percentage of funds not sent to the RiskFund contract when the funds are released, following the project Tokenomics
    uint256 private constant protocolSharePercentage = 70;
```

contracts/Comptroller.sol
```diff
@@ -1019,6 +1011,14 @@ contract Comptroller is Ownable2StepUpgradeable, ComptrollerV1Storage, Comptroll
 
     /*** Liquidity/Liquidation Calculations ***/
 
+    /**
```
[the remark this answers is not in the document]
UrAvgDeveloper: Not sure I understand which comment you are referring to @kkirka 🤔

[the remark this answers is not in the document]
kkirka: Ah, ok, it's been removed in the consecutive commits

contracts/BaseJumpRateModelV2.sol
chechu: We'll finally remove the require statement because the division by zero will generate a revert (we are using solidity => 0.8.0), so we don't need it

chechu (approved):
LGTM

defcon022 (approved):
LGTM!

PR #163, opened 2023-02-13
[VEN-963, VEN-967] Certik audit 

defcon022:
## Description
* 963 and 967 Certik audit fix.

Resolves #

## Checklist

### DIFF
--- a/contracts/ComptrollerStorage.sol
+++ b/contracts/ComptrollerStorage.sol
@@ -116,6 +116,13 @@ contract ComptrollerStorage {
     /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
     address public immutable accessControl;
 
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[50] private __gap;
+
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor(address poolRegistry_, address accessControl_) {
         // Note that the contract is upgradeable. We only initialize immutables in the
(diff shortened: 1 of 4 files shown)


contracts/Rewards/RewardsDistributor.sol
```diff
@@ -256,9 +256,7 @@ contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable {
      * @param vTokens The list of markets to claim REWARD TOKEN in
      */
     function claimRewardToken(address holder, VToken[] memory vTokens) public {
-        address[] memory holders = new address[](1);
-        holders[0] = holder;
-        _claimRewardToken(holders, vTokens);
+        _claimRewardToken(holder, vTokens);
```
chechu, line 259: Do we need the internal function? Would it make sense just move the function content of `_claimRewardToken` to this public function? We are not invoking `_claimRewardToken` from any other place

defcon022: cool, will move.

defcon022: [FIXED](https://github.com/VenusProtocol/isolated-pools/pull/163/commits/b3d53030012993270beaf7fa074763c64e5f7645)

0xlucian (approved):

contracts/ComptrollerStorage.sol
```diff
@@ -104,10 +104,34 @@ contract ComptrollerStorage {
...
+
     /**
      * @dev This empty reserved space is put in place to allow future versions to add new
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
     uint256[50] private __gap;
+
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor(address poolRegistry_, address accessControl_) {
```
narayanprusty, line 127: @defcon022 Shouldn't the constructor be in `Comptroller` contract instead of `ComptrollerStorage`?

defcon022: `ComptrollerStorage` is inherited in `Comptroller` and every storage variable is been shifted to `ComptrollerStorage` contract. 

The same issue been raised by Certik.

kkirka (approved):

kkirka (commented):
I personally don't think immutables belong in the storage contract. These are like constants that are put to code, not to the storage slots of the contract, so they are kinda like implementation. But I don't want to step in the way here.

PR #180, opened 2023-03-08
[VEN-953] Enhance access control mechanisms

kkirka:
This PR refactors the access control mechanism to use `AccessControlled` mixin and adds access controls to the following methods:

Shortfall:
  * startAuction
  * restartAuction

BaseJumpRateModelV2:
  * updateJumpRateModel

PoolRegistry:
  * addMarket
  * createRegistryPool
  * updatePoolMetadata
  * setPoolName

RewardsDistributor:
  * setRewardTokenSpeeds

RiskFund:
  * setMinAmountToConvert

Comptroller:
  * setCloseFactor

### DIFF
--- a/contracts/BaseJumpRateModelV2.sol
+++ b/contracts/BaseJumpRateModelV2.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 pragma solidity 0.8.13;
 
+import "./Governance/IAccessControlManager.sol";
 import "./InterestRateModel.sol";
 
 /**
@@ -12,9 +13,9 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
     uint256 private constant BASE = 1e18;
 
     /**
-     * @notice The address of the owner, i.e. the Timelock contract, which can update parameters directly
+     * @notice The address of the AccessControlManager contract
      */
-    address public owner;
+    IAccessControlManager public accessControlManager;
 
     /**
      * @notice The approximate number of blocks per year that is assumed by the interest rate model
@@ -41,6 +42,11 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
      */
     uint256 public kink;
 
+    /**
+     * @notice Thrown when the action is prohibited by AccessControlManager
+     */
+    error Unauthorized(address sender, address calledContract, string methodSignature);
+
     event NewInterestParams(
         uint256 baseRatePerBlock,
         uint256 multiplierPerBlock,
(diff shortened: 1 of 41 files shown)


chechu (commented):
LGTM

0xlucian (approved):

contracts/Governance/AccessControlled.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.13;
+
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+
+import "./IAccessControlManager.sol";
+
+abstract contract AccessControlled is Initializable, Ownable2StepUpgradeable {
```
chechu, line 9: Would it make sense to move (not now, in the future) this abstract class to the new Governance package?

kkirka: Makes sense IMO

UrAvgDeveloper (approved):
LGTM !

contracts/Governance/AccessControlled.sol
chechu: VEN-1213 created

contracts/BaseJumpRateModelV2.sol
```diff
@@ -12,9 +13,9 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
     uint256 private constant BASE = 1e18;
 
     /**
-     * @notice The address of the owner, i.e. the Timelock contract, which can update parameters directly
+     * @notice The address of the AccessControlManager contract
      */
-    address public owner;
+    IAccessControlManager public accessControlManager;
```
chechu, line 18: Would it make sense to extend `AccessControlled`, to reuse the definitions we already have in that abstract class?

chechu (approved):

contracts/BaseJumpRateModelV2.sol
kkirka: It's not easy actually, because:
1) AccessControlled mixin is designed for upgradeable contracts, i.e. it has an initializer method and inherits Ownable2StepUpgradeable, rather than having a constructor and inheriting Ownable2Step.
2) JumpRateModels are created by factories, so the default owner would be the factory. We'll need some custom logic to transfer the ownership then.

defcon022 (approved):
Awesome!

PR #181, opened 2023-03-13
Hacken Audit Fixes

narayanprusty:
## Description

**VEN-1204: L02. Redundant Import**
The RiskFund contract imports "../Pool/PoolRegistry.sol", "../Pool/PoolRegistryInterface.sol".
Recommendation: Remove the redundant import.

**VEN-1205: L08. Functions that Can Be Declared External**
Recommendation: Use the external attribute for functions never called from the contract.
Affected function: VToken.initialize()

## Checklist

- [x] I have updated the documentation to account for the changes in the code.
- [x] If I added new functionality, I added tests covering it.
- [x] If I fixed a bug, I added a test preventing this bug from silently reappearing again.
- [x] My contribution follows [Venus contribution guidelines](docs/CONTRIBUTING.md).

PR #183, opened 2023-03-15
[VEN-1172] Fix/int rates

0xlucian:
## Description

This PR adds reserveFactor to vToken constructor and adjusts the deployment config. It also chagnes `BaseJumpRateModelV2.blocksPerYear ` value with the correct one for BSC 

Resolves # VEN-1172

## Checklist


chechu (commented):
Reviewing the code... why don't we simply invoke the function `setReserveFactor` on the VToken contract deployed after deploying it? We would avoid this change in the contracts, and it would be the same approach we could follow to set other market parameters, like `protocolSeizeShareMantissa`

0xlucian:
the reason for that is because `setReserveFactor` is access controlled and we need to give deployer default access to this function. If deployer account gets compromised at some point that might be a big problem. Also on the other hand it really makes logic of deployment scripts more complicated... Setting this in the initialise function we call the `_setReserveFactorFresh` which is internal and not guarded... I think that we actually discussed this possible implementation (only change in the deployment script) but we agreed on changing the smart contract logic for `reserveFactorMantissa` initialisation

contracts/test/VTokenHarness.sol
```diff
@@ -12,6 +12,35 @@ contract VTokenHarness is VToken {
 
     mapping(address => bool) public failTransferToAddresses;
 
+    function initializeHarness(
```
chechu, line 15: We removed these items [here](https://github.com/VenusProtocol/isolated-pools/pull/181/files#diff-35683b5d7bfaed24c286787c300ea5bf1baf9fd0fc3c12ad5a019d6a56908d51). Do we need them?

narayanprusty: I am calling the `initialise` directly from tests without a wrapper. @0xlucian so during rebase you can do the same. Should be straightforward.

chechu (approved):
LGTM

narayanprusty (approved):

defcon022 (approved):
Looks good.

PR #190, opened 2023-03-28
[VEN-1245] feat: make incentiveBps configurable

coreyar:
## Description

Make incentiveBps configurable

Resolves VEN-1245

### DIFF
--- a/contracts/Shortfall/Shortfall.sol
+++ b/contracts/Shortfall/Shortfall.sol
@@ -320,7 +320,8 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
      * @custom:event Emits MinimumPoolBadDebtUpdated on success
      * @custom:access Restricted to owner
      */
-    function updateMinimumPoolBadDebt(uint256 _minimumPoolBadDebt) external onlyOwner {
+    function updateMinimumPoolBadDebt(uint256 _minimumPoolBadDebt) external {
+        _checkAccessAllowed("updateMinimumPoolBadDebt(uint256)");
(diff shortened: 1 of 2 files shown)


contracts/Shortfall/Shortfall.sol
```diff
@@ -279,6 +283,19 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
         _startAuction(comptroller);
     }
 
+    /**
+     * @notice Updates the inventive BPS
+     * @param _incentiveBps New incentive BPS
+     * @custom:event Emits IncentiveBpsUpdated on success
+     * @custom:access Restricted to owner
+     */
+    function updateIncentiveBps(uint256 _incentiveBps) external onlyOwner {
```
chechu, line 292: I would integrate this with the ACM instead of using `onlyOwner`

contracts/Shortfall/Shortfall.sol
```diff
@@ -279,6 +283,19 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
...
+        require(_incentiveBps != 0, "incentiveBps must not be 0");
```
chechu, line 293: Do we know an upper limit? 100%?

contracts/Shortfall/Shortfall.sol
coreyar: I've integrated this with ACM. I also added commits to put the other setter functions behind the ACM to avoid another PR and conflicts

aditya-routv (approved):

chechu (approved):

PR #197, opened 2023-04-05
feat: add ability to toggle auctions on and off

coreyar:
## Description

Adds toggle to turn auctions on and off.

### DIFF
--- a/contracts/Shortfall/Shortfall.sol
+++ b/contracts/Shortfall/Shortfall.sol
@@ -71,6 +71,9 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
     /// @notice Time to wait for next bidder. initially waits for 10 blocks
     uint256 public nextBidderBlockLimit;
 
+    /// @notice Boolean of if auctions are paused
+    bool public auctionsPaused;
+
     /// @notice Time to wait for first bidder. initially waits for 100 blocks
     uint256 public waitForFirstBidder;
 
@@ -123,6 +126,12 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
     /// @notice Emitted when incentiveBps is updated
     event IncentiveBpsUpdated(uint256 oldIncentiveBps, uint256 newIncentiveBps);
 
+    /// @notice Emitted when auctions are paused
+    event AuctionsPaused(address sender);
+
+    /// @notice Emitted when auctions are unpaused
+    event AuctionsResumed(address sender);
+
     /// @custom:oz-upgrades-unsafe-allow constructor
     constructor() {
         // Note that the contract is upgradeable. Use initialize() or reinitializers
@@ -158,6 +167,7 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
         waitForFirstBidder = DEFAULT_WAIT_FOR_FIRST_BIDDER;
         nextBidderBlockLimit = DEFAULT_NEXT_BIDDER_BLOCK_LIMIT;
         incentiveBps = DEFAULT_INCENTIVE_BPS;
+        auctionsPaused = false;
     }
 
     /**
@@ -216,6 +226,7 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
      * @notice Close an auction
      * @param comptroller Comptroller address of the pool
      * @custom:event Emits AuctionClosed event on successful close
+     * @custom:event Errors if auctions are paused
      */
     function closeAuction(address comptroller) external nonReentrant {
         Auction storage auction = auctions[comptroller];
@@ -273,8 +284,10 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
      * @notice Start a auction when there is not currently one active
      * @param comptroller Comptroller address of the pool
      * @custom:event Emits AuctionStarted event on success
+     * @custom:event Errors if auctions are paused
      */
     function startAuction(address comptroller) external {
+        require(!auctionsPaused, "Auctions are paused");
         _startAuction(comptroller);
     }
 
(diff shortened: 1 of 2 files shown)


chechu (commented):
We could follow an approach similar to what we have done in the Vault contracts [here](https://github.com/VenusProtocol/venus-protocol/pull/236), making them "pausable". It uses a modifier and define the guard variable as "pause" (opposite to "enabled")

chechu:
> We could follow an approach similar to what we have done in the Vault contracts [here](https://github.com/VenusProtocol/venus-protocol/pull/236), making them "pausable". It uses a modifier and defines the guard variable as "pause" (opposite to "enabled")

And we could extend this "pause" barrier to other functions: placeBid, closeAuction, restartAuction

coreyar:
> > We could follow an approach similar to what we have done in the Vault contracts [here](https://github.com/VenusProtocol/venus-protocol/pull/236), making them "pausable". It uses a modifier and defines the guard variable as "pause" (opposite to "enabled")
> 
> And we could extend this "pause" barrier to other functions: placeBid, closeAuction, restartAuction

I thought about this and decided that it isn't necessary to "pause" those functions.

If an auction is currently in progress we should let it finish which means bid and close still need to work if auctions are paused. Once auctions are paused `placeBid` and `closeAuction` will no longer work because there will not be any auctions. `restartAuction` won't work because it calls `startAuction`.

If we want to stop in progress auctions we need to add extra logic to send the locked funds back and end the auction

coreyar:
> > > We could follow an approach similar to what we have done in the Vault contracts [here](https://github.com/VenusProtocol/venus-protocol/pull/236), making them "pausable". It uses a modifier and defines the guard variable as "pause" (opposite to "enabled")
> > 
> > 
> > And we could extend this "pause" barrier to other functions: placeBid, closeAuction, restartAuction
> 
> I thought about this and decided that it isn't necessary to "pause" those functions.
> 
> If an auction is currently in progress we should let it finish which means bid and close still need to work if auctions are paused. Once auctions are paused `placeBid` and `closeAuction` will no longer work because there will not be any auctions. `restartAuction` won't work because it calls `startAuction`.
> 
> If we want to pause in progress auctions, I think we need to add extra logic to send the locked funds back and end the auction because it isn't right for us to lock those funds until the auctions are unpaused again.

Auctions will be paused and unpaused via VIP so participants will have time to react before pausing.

contracts/Shortfall/Shortfall.sol
```diff
@@ -354,6 +371,25 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
...
+     * @custom:access Restricted by ACM
+     */
+    function pauseAuctions() external {
+        _checkAccessAllowed("pauseAuctions()");
+        require(auctionsPaused == false, "Auctions are already paused");
+        auctionsPaused = true;
+        emit AuctionsPaused(msg.sender);
+    }
+
+    function resumeAuctions() external {
```
GitGuru7, line 386: Likely above we can add netspec comments in this external function

defcon022 (commented):
Tests are failing.

contracts/Shortfall/Shortfall.sol
```diff
@@ -354,6 +371,25 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
...
     }
 
+    /**
+     * @notice Toggle whether auctions are enabled or disabled
+     * @custom:event Emits AuctionEnabledToggled on success
+     * @custom:access Restricted by ACM
+     */
+    function pauseAuctions() external {
+        _checkAccessAllowed("pauseAuctions()");
+        require(auctionsPaused == false, "Auctions are already paused");
```
chechu, line 381:
```suggestion
        require(!auctionsPaused, "Auctions are already paused");
```

contracts/Shortfall/Shortfall.sol
```diff
@@ -354,6 +371,25 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
...
+        auctionsPaused = true;
+        emit AuctionsPaused(msg.sender);
+    }
+
+    function resumeAuctions() external {
+        _checkAccessAllowed("resumeAuctions()");
+        require(auctionsPaused == true, "Auctions are not paused");
```
chechu, line 388:
```suggestion
        require(auctionsPaused, "Auctions are not paused");
```

contracts/Shortfall/Shortfall.sol
```diff
@@ -261,15 +275,18 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlled, ReentrancyGuard
      * @notice Start a auction when there is not currently one active
      * @param comptroller Comptroller address of the pool
      * @custom:event Emits AuctionStarted event on success
+     * @custom:event Errors if auctions are paused
      */
     function startAuction(address comptroller) external {
+        require(auctionsPaused == false, "Auctions are paused");
```
chechu, line 281:
```suggestion
        require(!auctionsPaused, "Auctions are paused");
```

contracts/Shortfall/Shortfall.sol
```diff
@@ -364,6 +378,32 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
         emit PoolRegistryUpdated(oldPoolRegistry, poolRegistry_);
     }
 
+    /**
+     * @notice Pause auctions. This disables starting new auctions but lets the current auction finish.s
```
chechu, line 382:
```suggestion
     * @notice Pause auctions. This disables starting new auctions but lets the current auction finishes
```

tests/hardhat/Shortfall.ts
```diff
@@ -614,6 +616,29 @@ describe("Shortfall: Tests", async function () {
...
+        .to.emit(shortfall, "AuctionsPaused")
+        .withArgs(owner.address);
+      // vDAI.badDebt.returns(parseUnits("10000", 18));
+      // await vDAI.setVariable("badDebt", parseUnits("10000", 18));
+      // vWBTC.badDebt.returns(parseUnits("2", 8));
+      // await vWBTC.setVariable("badDebt", parseUnits("2", 8));
+
+      // await expect(shortfall.startAuction(poolAddress)).to.be.revertedWith("Auctions are paused");
+      // await expect(shortfall.closeAuction(poolAddress)).to.be.revertedWith("Auctions are paused");
+      // await expect(shortfall.placeBid(poolAddress, 1)).to.be.revertedWith("Auctions are paused");
```
chechu, line 641: ?

chechu (approved):

PR #206, opened 2023-04-27
[VEN-1236] Feat/ Semantic Release

coreyar:
## Description

Add semantic release for automating versioning and releasing packages


.yarn/releases/yarn-1.22.1.cjs
chechu, line 1: Do we need this file?

chechu (approved):

.yarn/releases/yarn-1.22.1.cjs
coreyar: I added it because our repos use different versions of yarn and switching node versions caused the yarn version to get crossed.
This repo uses yarn classsic while others use yarn 3

chechu (approved):

contracts/Governance/AccessControlled.sol
```diff
@@ -50,6 +41,15 @@ abstract contract AccessControlled is Initializable, Ownable2StepUpgradeable {
...
     }
 
+    function __AccessControlled_init(address accessControlManager_) internal onlyInitializing {
+        __Ownable2Step_init();
+        __AccessControlled_init_unchained(accessControlManager_);
+    }
+
+    function __AccessControlled_init_unchained(address accessControlManager_) internal onlyInitializing {
+        _setAccessControlManager(accessControlManager_);
+    }
```
kkirka, line 51: Initializer methods are much like constructors, so even though some initializers are internal IMO it makes more sense to put them to the beginning of the contracts. However, if it breaks linters, I'm fine with putting them to internal functions block either. Can we settle on a single approach so that we don't go back and forth with this?

kkirka (approved):

PR #207, opened 2023-04-28
[VEN-1151][VEN-1152][VEN-2046] send reserves to PSR  and remove PSR contracts

GitGuru7:
## Description

This PR will add feature to move reserves fund from vToken to protocolShareReserve

Resolves [VEN-1151][VEN-1152]

## Checklist


contracts/VToken.sol
```diff
@@ -1111,22 +1142,31 @@ contract VToken is Ownable2StepUpgradeable, AccessControlled, VTokenInterface, E
...
+            address(comptroller),
+            underlying,
+            IProtocolShareReserve.IncomeType.LIQUIDATION
+        );
+
         /* Emit a Transfer event */
         emit Transfer(borrower, liquidator, liquidatorSeizeTokens);
         emit Transfer(borrower, address(this), protocolSeizeTokens);
-        emit ReservesAdded(address(this), protocolSeizeAmount, totalReservesNew);
+        emit Transfer(address(this), protocolShareReserve, protocolSeizeAmount);
```
chechu, line 1169: I think we can merge these two events into only 1, transferring from the borrower to the `protocolShareReserve` contract. wdyt?

contracts/VToken.sol
```diff
@@ -330,15 +337,22 @@ contract VToken is Ownable2StepUpgradeable, AccessControlled, VTokenInterface, E
...
      */
-    function reduceReserves(uint256 reduceAmount) external override nonReentrant {
+    function reduceReserves(uint256 reduceAmount, IProtocolShareReserve.IncomeType kind)
+        external
+        override
+        nonReentrant
+    {
         accrueInterest();
-        _reduceReservesFresh(reduceAmount);
+        if (reduceReservesBlockNumber == _getBlockNumber()) return;
```
chechu, line 354: Why do we have this check?

contracts/RiskFund/IProtocolShareReserve.sol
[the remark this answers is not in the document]
GitGuru7: ACK, Currently we can ignore the changes made in this risk related contracts

contracts/VToken.sol
GitGuru7: If reserves are already reduced in `accrueInterest` we should fail it gracefully that's why we need this check

contracts/VToken.sol
GitGuru7: There are actually three transfers taking place in this operation, and if we combine these two, we will lose track of how many tokens from the borrower has been seized of.

contracts/RiskFund/ReserveHelpers.sol
```diff
@@ -19,11 +20,19 @@ contract ReserveHelpers {
...
+    mapping(address => mapping(address => uint256)) internal poolsLiquidationReserves;
+
+    mapping(address => mapping(address => uint256)) internal poolsSpreadReserves;
+
     /**
      * @dev This empty reserved space is put in place to allow future versions to add new
      * variables without shifting down storage in the inheritance chain.
      */
-    uint256[48] private __gap;
+    uint256[44] private __gap;
```
chechu, line 35: I think we are not using and we don't need these changes

contracts/RiskFund/ProtocolShareReserve.sol
chechu, line 1: Let's start using the new package `@venusprotocol/protocol-reserve`, version `^1.0.0-dev.2`. By doing it we should avoid changes in the `contracts/RiskFund` folder in this PR

contracts/VToken.sol
```diff
@@ -5,14 +5,15 @@ import { Ownable2StepUpgradeable } from "@openzeppelin/contracts-upgradeable/acc
...
+import { IProtocolShareReserve } from "@venusprotocol/protocol-reserve/contracts/Interfaces/IProtocolShareReserve.sol";
 
 import { VTokenInterface } from "./VTokenInterfaces.sol";
 import { ComptrollerInterface, ComptrollerViewInterface } from "./ComptrollerInterface.sol";
 import { TokenErrorReporter } from "./ErrorReporter.sol";
 import { InterestRateModel } from "./InterestRateModel.sol";
 import { ExponentialNoError } from "./ExponentialNoError.sol";
-import { IProtocolShareReserve } from "./RiskFund/IProtocolShareReserve.sol";
 import { ensureNonzeroAddress } from "./lib/validators.sol";
+import { AccessControlledV8 } from "@venusprotocol/governance-contracts/contracts/Governance/AccessControlledV8.sol";
```
chechu, line 16: Duplicated

contracts/VToken.sol
```diff
@@ -781,6 +799,11 @@ contract VToken is
         totalBorrows = totalBorrowsNew;
         totalReserves = totalReservesNew;
 
+        if (currentBlockNumber - reduceReservesBlockNumber >= reduceReservesBlockDelta) {
```
kkirka, line 802: I wonder what happens if we forget to initialize this variable. Will the reserves be withdrawn every block?

Maybe we'd rather want to treat 0 as "automatic allocation is switched off"?

```suggestion
        uint256 reduceReservesBlockDelta_ = reduceReservesBlockDelta;
        if (reduceReservesBlockDelta_ && currentBlockNumber - reduceReservesBlockNumber >= reduceReservesBlockDelta) {
```

kkirka (approved):

kkirka:
@GitGuru7 maybe we want to include this `reduceReservesBlockDelta` to `AddMarketInput` so that we don't accidentally forget to set this value?

contracts/VToken.sol
GitGuru7: @kkirka I believe that initially we won't be missing it anyway because we will add all of the necessary initialization in the VToken upgrades https://github.com/VenusProtocol/vips/pull/67.

GitGuru7:
> @GitGuru7 maybe we want to include this `reduceReservesBlockDelta` to `AddMarketInput` so that we don't accidentally forget to set this value?

If you're referring to the PoolRegistry addMarket input, we can certainly add it so that we don't have to initialise it each time we add a new market in the VIP, but I'd prefer to keep it outside the scope of this ticket.

cc: @chechu

chechu:
> > @GitGuru7 maybe we want to include this `reduceReservesBlockDelta` to `AddMarketInput` so that we don't accidentally forget to set this value?
> 
> @kkirka if you're referring to the PoolRegistry addMarket input, we can certainly add it so that we don't have to initialise it each time we add a new market in the VIP, but I'd prefer to keep it outside the scope of this ticket WDYT?
> 
> cc: @chechu

We have VEN-2094 for it. Out of the scope for this PR

contracts/VTokenInterfaces.sol
```diff
@@ -271,6 +281,13 @@ abstract contract VTokenInterface is VTokenStorage {
      */
     event SweepToken(address indexed token);
```
chechu, line 283: Missing natspec comment

contracts/VToken.sol
```diff
@@ -1226,8 +1257,12 @@ contract VToken is
      * @notice Reduces reserves by transferring to the protocol reserve contract
      * @dev Requires fresh interest accrual
      * @param reduceAmount Amount of reduction to reserves
+     * @param kind kind of reserve
      */
-    function _reduceReservesFresh(uint256 reduceAmount) internal {
+    function _reduceReservesFresh(uint256 reduceAmount, IProtocolShareReserve.IncomeType kind) internal {
```
chechu, line 1262: `kind` is always `IProtocolShareReserve.IncomeType.SPREAD`. I would remove the param and I would hardcode the enum value in this internal function.

chechu (approved):

PR #244, opened 2023-06-12
[VEN-979]: add natspec comments at contract levels and fix the template

chechu:
## Description

Related to VEN-979 (Isolated pools). It adds doc at the contract level, and fixes the `page.hbs` template 

## Checklist

### DIFF
--- a/docgen-templates/page.hbs
+++ b/docgen-templates/page.hbs
@@ -1,5 +1,5 @@
 {{h}} {{items.0.natspec.title}}
-{{items.0.natspec.notice}}
+{{{items.0.natspec.notice}}}
 
 # Solidity API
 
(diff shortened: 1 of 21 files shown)


docgen-templates/page.hbs
```diff
@@ -1,9 +1,10 @@
 {{h}} {{items.0.natspec.title}}
-{{items.0.natspec.notice}}
+{{{items.0.natspec.notice}}}
```
coreyar, line 2: I don't think this should have three brackets. I think it should have 2

chechu: Needed to avoid the escape of characters like `

coreyar (approved):

PR #245, opened 2023-06-13
chore: modify deployment scripts to support VIP-based deployment

kkirka:
This PR modifies deployment scrips to:
1. First, deploy the Comptrollers, VTokens, and RewardsDistributors
2. Then transfer ownership and, potentially, faucet and approve initial supply to deployer or timelock, depending on the network
3. Then prepare a VIP or execute the VIP commands (if launched in `hardhat` network).

I tried to mimic the existing logic that would allow running the scripts once again with a different config (it would mean "create and configure the missing pools, markets, rewards").


deploy/009-deploy-vtokens.ts
```diff
@@ -0,0 +1,134 @@
...
+        tokenContract.address,
+        comptrollerProxy.address,
+        rateModelAddress,
+        parseUnits("1", underlyingDecimals + 18 - vTokenDecimals),
+        name,
+        symbol,
+        vTokenDecimals,
+        preconfiguredAddresses.NormalTimelock || deployer, // admin
+        accessControlManagerAddress,
+        [treasuryAddress, ADDRESS_ONE],
```
chechu, line 117: Give this definition of the `RiskManagementInit`:
```
    struct RiskManagementInit {
        address shortfall;
        address payable protocolShareReserve;
    }
```
Shouldn't we change the order of the params, setting `[ADDRESS_ONE, treasuryAddress]`, so the Shortfall will be `ADDRESS_ONE` and the protocolShareReserve the treasury?

deploy/011-initial-liquidity.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const getUnderlyingToken = async (assetSymbol: string, tokensConfig: TokenConfig[]): Promise<ERC20> => {
+  const token = getTokenConfig(assetSymbol, tokensConfig);
+  let underlyingAddress = token.tokenAddress;
+  if (token.isMock) {
+    underlyingAddress = (await getUnderlyingMock(assetSymbol)).address;
+  }
+  return ethers.getContractAt<ERC20>("@openzeppelin/contracts/token/ERC20/ERC20.sol:ERC20", underlyingAddress);
+};
```
chechu, line 20: These two functions are already defined (in a very similar way) and used in 013. Just in case you would like to consider moving them to the deploymentUtils.ts file :)

deploy/012-transfer-pools-ownership.ts
```diff
@@ -0,0 +1,73 @@
...
+      await tx.wait(1);
+      const newOwner = await contract.owner();
+      console.log(`${contractName} owner ${owner} sucessfully changed to ${newOwner}.`);
+    } else {
+      console.error(`${contractName} owner ${owner} is equal to target ownership address ${targetOwner}`);
+    }
+  }
+};
+
+func.tags = ["TransferPoolsOwnership"];
```
chechu, line 72: Do we need the tag `"il"`?

helpers/deploymentUtils.ts
```diff
@@ -0,0 +1,119 @@
...
+export const getUnregisteredPools = async (
+  poolConfig: PoolConfig[],
+  hre: HardhatRuntimeEnvironment,
+): Promise<PoolConfig[]> => {
+  const { deployments } = hre;
+  const registry = await ethers.getContract("PoolRegistry");
+  const registeredPools = (await registry.getAllPools()).map((p: { comptroller: string }) => p.comptroller);
+  const isRegistered = await Promise.all(
+    poolConfig.map(async pool => {
+      const comptroller = await deployments.getOrNull(`Comptroller_${pool.name}`);
```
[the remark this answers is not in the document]
kkirka: I just need an address (or null) here, I don't need an ethers contract. Maybe we could just remove `getPoolComptroller`?

PR #246, opened 2023-06-13
[VEN-1567]: add sweep token in ReserveHelpers

GitGuru7:
## Description

This PR add `SweepToken` function in the `ReserverHelpers`. Allow only sweeping the tokens not tracked in `assetsReserves`, or the surplus (untracked) in these tokens. Send the tokens to the admin/owner of the contract, to avoid potential issues.

Resolves VEN-1567

## Checklist


contracts/RiskFund/ReserveHelpers.sol
```diff
@@ -33,6 +43,40 @@ contract ReserveHelpers {
...
+     * @custom:access Only Owner
+     */
+    function sweepToken(address _token, address _to) external onlyOwner nonReentrant {
+        ensureNonzeroAddress(_to);
+        uint256 balanceDfference_;
+        uint256 balance_ = IERC20Upgradeable(_token).balanceOf(address(this));
+
+        require(balance_ > assetsReserves[_token], "ReserveHelpers: Zero surplus tokens");
+        unchecked {
+            balanceDfference_ = balance_ - assetsReserves[_token];
```
defcon022, line 73: After transferring funds by mistake, if `updateAssetsState` is called either at the time of protocol share transfer / vToken or manually since it's a public function then we can not sweepTokens for the user. 
I have two suggestions:
1. We can check for the `isVToken` and also we have to implement a similar for `ProtocolShareReserve` like `isProtocolShareReserve` while calling `updateAssetsState` so that this function can only be called while transferring funds from these contracts.
2. we can pass the amount to be swept as this is an onlyOwner function and directly transfer the amount to the user without any checks.

defcon022:
@GitGuru7 Please resolve the conflicts.

contracts/RiskFund/ReserveHelpers.sol
GitGuru7: AFAIK tokens should be transfer to admin/owner of the contract to avoid any potential issues acc to ticket.
@chechu I think you can provide more info here

chechu: We are already doing something similar to the first checked mentioned by @Debugger022 here: https://github.com/VenusProtocol/isolated-pools/blob/develop/contracts/RiskFund/ReserveHelpers.sol#L66

During the execution of `updateAssetsState` we check that the asset is associated with a valid market in the Comptroller. So, I don't think we need any extra check

kkirka (approved):

contracts/RiskFund/ReserveHelpers.sol
Debugger022: @chechu @GitGuru7 
Reiterating my point:

If a user mistakenly sends funds to the contract and the `updateAssetsState` function is called before executing the `sweepToken` function, the user's funds will be considered part of the reserves, and they will be unable to withdraw their funds in any way.

chechu: I prefer to accept that case and limit the power of the owner to sweep arbitrary amounts. So, I wouldn't change it. Users shouldn't interact directly with this contract, so I hope no one really sends any tokens to it.

Debugger022 (approved):

PR #285, opened 2023-08-07
[VEN-1812]: Peckshield audit fixes in automatic income allocation 

GitGuru7:
## Description
- PVE006 -> Change event names

## Checklist


contracts/VToken.sol
```diff
@@ -1223,8 +1223,7 @@ contract VToken is
 
         /* Emit a Transfer event */
         emit Transfer(borrower, liquidator, liquidatorSeizeTokens);
-        emit Transfer(borrower, address(this), protocolSeizeTokens);
-        emit Transfer(address(this), protocolShareReserve, protocolSeizeAmount);
+        emit LiquidationReservesReduced(borrower, protocolShareReserve, protocolSeizeAmount);
```
chechu, line 1226: I would use a different name, because we are not reducing anything here. If we transfer the income generated during liquidations every time that there is a liquidation, we don't generate "liquidation reserves". What about `ProtocolSeize`?

GitGuru7: fixed here [4576435](https://github.com/VenusProtocol/isolated-pools/pull/285/commits/45764358d69e3afe39b990d1e7b181ae1023be45)

chechu (approved):

chechu (approved):

PR #287, opened 2023-08-09
[VEN-1798]: Fix Certik 2023-08 findings for PSR/RiskFund/Shortfall contracts

kkirka:
This PR fixes the findings identified by Certik in their 2023-08 audit of PSR, RiskFund and Shortfall contracts.

- [x] VPB-05 | Typos and Inconsistencies
- [ ] (acknowledged) SSV-03 | Potential To Fill Blocks To Stop Others Bids To Win Auction
- [ ] (TBD) VPB-01 | Custom Errors Can Be Used
- [ ] (TBD) TDT-01 | Unchecked Blocks Can Optimize Contract
- [x] SSV-06 | Unnecessary Update to `marketDebt` in `_startAuction()`
- [x] SSV-04 | Check on Auction Status Can Be Simplified
- [x] RHR-01 | View Function Equivalent to Compiler-Generated Getters
- [x] RFR-01 | Redundant Check on `amountOutMin`
- [x] VPB-04 | Ineffectual Use of Reentrancy Guard
- [x] RFR-03 | Issue With `_swapAsset()` Depending On `convertibleBaseAsset` Price
- [x] SSV-01 | `_transferOutOrTrackDebt()` Is Not Used When Closing Auction

### DIFF
--- a/contracts/RiskFund/RiskFund.sol
+++ b/contracts/RiskFund/RiskFund.sol
@@ -18,7 +18,7 @@ import { MaxLoopsLimitHelper } from "../MaxLoopsLimitHelper.sol";
 import { ensureNonzeroAddress } from "../lib/validators.sol";
 
 /**
- * @title ReserveHelpers
+ * @title RiskFund
  * @author Venus
  * @notice Contract with basic features to track/hold different assets for different Comptrollers.
  * @dev This contract does not support BNB.
--- a/contracts/RiskFund/RiskFund.sol
+++ b/contracts/RiskFund/RiskFund.sol
@@ -156,7 +156,7 @@ contract RiskFund is AccessControlledV8, ExponentialNoError, ReserveHelpers, Max
         address[][] calldata paths,
         uint256 deadline
     ) external override nonReentrant returns (uint256) {
-        _checkAccessAllowed("swapPoolsAssets(address[],uint256[],address[][])");
+        _checkAccessAllowed("swapPoolsAssets(address[],uint256[],address[][],uint256)");
         require(deadline >= block.timestamp, "Risk fund: deadline passed");
         address poolRegistry_ = poolRegistry;
         ensureNonzeroAddress(poolRegistry_);
--- a/contracts/RiskFund/RiskFund.sol
+++ b/contracts/RiskFund/RiskFund.sol
@@ -4,6 +4,7 @@ pragma solidity 0.8.13;
 import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import { SafeERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
 import { AccessControlledV8 } from "@venusprotocol/governance-contracts/contracts/Governance/AccessControlledV8.sol";
+import { ResilientOracleInterface } from "@venusprotocol/oracle/contracts/interfaces/OracleInterface.sol";
 import { ComptrollerInterface } from "../ComptrollerInterface.sol";
 import { IRiskFund } from "./IRiskFund.sol";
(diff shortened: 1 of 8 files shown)


contracts/RiskFund/RiskFund.sol
```diff
@@ -233,7 +233,7 @@ contract RiskFund is AccessControlledV8, ExponentialNoError, ReserveHelpers, Max
      */
     function getPoolsBaseAssetReserves(address comptroller) external view returns (uint256) {
         require(ComptrollerInterface(comptroller).isComptroller(), "Risk Fund: Comptroller address invalid");
-        return poolsAssetsReserves[comptroller][convertibleBaseAsset];
+        return _poolsAssetsReserves[comptroller][convertibleBaseAsset];
     }
```
chechu, line 238: Would it make sense to have something similar for the ProtocolShareReserves contract? In that, could we move this  function to the ReserveHelpers file?

contracts/RiskFund/RiskFund.sol
```diff
@@ -260,48 +261,47 @@ contract RiskFund is AccessControlledV8, ExponentialNoError, ReserveHelpers, Max
...
-        uint256 underlyingAssetPrice = ComptrollerViewInterface(comptroller).oracle().getUnderlyingPrice(
-            address(vToken)
-        );
+        ResilientOracleInterface oracle = ComptrollerViewInterface(comptroller).oracle();
+        oracle.updateAssetPrice(convertibleBaseAsset_);
+        Exp memory  baseAssetPrice = Exp({ mantissa: oracle.getPrice(convertibleBaseAsset_) });
+        uint256 amountOutMinInUsd = mul_ScalarTruncate(baseAssetPrice, amountOutMin);
+
+        if (amountOutMinInUsd < minAmountToConvert) {
+            return 0;
```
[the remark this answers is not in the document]
kkirka: I think this was the idea, yeah. I didn't change the behavior in this regard. But I think reverting should be fine as well, maybe it would better match the expectations of the callers (e.g. imagine the oracle price has changed – we'd probably want to fail rather than "do other swaps and let the user check which ones were indeed executed"). LMK what you think.

GitGuru7 (dismissed):

contracts/RiskFund/RiskFund.sol
kkirka: There's no notion of "base asset" in reserve helpers, so I think we should be fine with `getPoolAssetReserve`. `getPoolsBaseAssetReserves(comptroller)` is basically `getPoolsBaseAssetReserves(comptroller, convertibleBaseAsset)`

kkirka: But we may need sth like this in the _interface_ though

contracts/RiskFund/RiskFund.sol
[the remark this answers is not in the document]
chechu: It's ok to return 0 if `balanceOfUnderlyingAsset == 0`, but I would stop the TX in the `minAmountToConvert` check if it's unsatisfied.

Initially, we this require:
```
require(amountOutMin >= minAmountToConvert, "RiskFund: amountOutMin should be greater than minAmountToConvert");
```

I know the units were wrong, but it was stopping the TX

contracts/RiskFund/RiskFund.sol
chechu: My fault, I didn't take into account `ReserveHelpers.getPoolAssetReserve`. We can keep it as it is

contracts/RiskFund/RiskFund.sol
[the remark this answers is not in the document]
kkirka: https://github.com/VenusProtocol/isolated-pools/pull/287/commits/9930fd5298ddc677992b28360fc6eb7931ff6dcd

chechu (approved):

chechu (dismissed):

PR #305, opened 2023-09-13
[VEN-1930]: forced liquidation

Debugger022:
## Description

Resolves #
- VEN-1930

## Checklist

### DIFF
--- a/contracts/Comptroller.sol
+++ b/contracts/Comptroller.sol
@@ -478,7 +478,7 @@ contract Comptroller is
         uint256 borrowBalance = VToken(vTokenBorrowed).borrowBalanceStored(borrower);
 
         /* Allow accounts to be liquidated if the market is deprecated or it is a forced liquidation */
-        if (skipLiquidityCheck || isDeprecated(VToken(vTokenBorrowed)) || isForcedLiquidationEnabled[vTokenBorrowed]) {
+        if (skipLiquidityCheck || isForcedLiquidationEnabled[vTokenBorrowed]) {
             if (repayAmount > borrowBalance) {
                 revert TooMuchRepay();
             }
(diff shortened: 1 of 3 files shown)


contracts/Comptroller.sol
```diff
@@ -475,7 +478,7 @@ contract Comptroller is
         uint256 borrowBalance = VToken(vTokenBorrowed).borrowBalanceStored(borrower);
 
         /* Allow accounts to be liquidated if the market is deprecated or it is a forced liquidation */
-        if (skipLiquidityCheck || isDeprecated(VToken(vTokenBorrowed))) {
+        if (skipLiquidityCheck || isDeprecated(VToken(vTokenBorrowed)) || isForcedLiquidationEnabled[vTokenBorrowed]) {
```
chechu, line 481: Let's remove the `isDeprecated` function. We are replacing it with the flag `isForcedLiquidationEnabled`, so we can enable/disable forced liquidations in a arbitrary way with a VIP

contracts/ComptrollerStorage.sol
```diff
@@ -118,10 +118,13 @@ contract ComptrollerStorage {
     // No collateralFactorMantissa may exceed this value
     uint256 internal constant MAX_COLLATERAL_FACTOR_MANTISSA = 0.9e18; // 0.9
 
+    /// @notice Flag indicating whether forced liquidation enabled for a market
+    mapping(address => bool) public isForcedLiquidationEnabled;
+
```
kkirka, line 123: Maybe we could put the new var before the constants block or move the constants block to the top of the file so that we have some separation between these?

tests/hardhat/Comptroller/liquidateAccountTest.ts
```diff
@@ -342,4 +343,129 @@ describe("liquidateAccount", () => {
...
+    });
+  });
+
+  describe("preLiquidateHook", async () => {
+    let accounts: SignerWithAddress[];
+
+    before(async () => {
+      accounts = await ethers.getSigners();
+    });
+    const generalTests = () => {
```
kkirka, line 388: We don't need this function if we don't have the second block. I guess we could just put these test cases right into the "isForcedLiquidationEnabled == true" block.

kkirka (commented):
LGTM, though I second the request to remove isDeprecated check

contracts/Comptroller.sol
Debugger022: Removed, cbd9b18a99c4e1f92bf9404e88fceb8ebc36d55f

contracts/ComptrollerStorage.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/305/commits/cbd9b18a99c4e1f92bf9404e88fceb8ebc36d55f

tests/hardhat/Comptroller/liquidateAccountTest.ts
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/305/commits/cbd9b18a99c4e1f92bf9404e88fceb8ebc36d55f

chechu (dismissed):

chechu (approved):

PR #319, opened 2023-10-30
[VEN-2107]: Quantstamp audit fixes on XVS bridge 

GitGuru7:
Resolves VEN-2107

## Checklist

### DIFF
--- a/contracts/Bridge/BaseXVSProxyOFT.sol
+++ b/contracts/Bridge/BaseXVSProxyOFT.sol
@@ -148,6 +148,7 @@ abstract contract BaseXVSProxyOFT is Pausable, ExponentialNoError, BaseOFTV2 {
      * @custom:event Emits SetMaxSingleTransactionLimit with old and new limit associated with chain id.
      */
     function setMaxSingleTransactionLimit(uint16 chainId_, uint256 limit_) external onlyOwner {
+        require(limit_ <= chainIdToMaxDailyLimit[chainId_], "Single transaction limit > Daily limit");
         emit SetMaxSingleTransactionLimit(chainId_, chainIdToMaxSingleTransactionLimit[chainId_], limit_);
         chainIdToMaxSingleTransactionLimit[chainId_] = limit_;
     }
@@ -173,6 +174,7 @@ abstract contract BaseXVSProxyOFT is Pausable, ExponentialNoError, BaseOFTV2 {
      * @custom:event Emits setMaxSingleReceiveTransactionLimit with old and new limit associated with chain id.
      */
     function setMaxSingleReceiveTransactionLimit(uint16 chainId_, uint256 limit_) external onlyOwner {
+        require(limit_ <= chainIdToMaxDailyReceiveLimit[chainId_], "single receive transaction limit > Daily limit");
         emit SetMaxSingleReceiveTransactionLimit(chainId_, chainIdToMaxSingleReceiveTransactionLimit[chainId_], limit_);
         chainIdToMaxSingleReceiveTransactionLimit[chainId_] = limit_;
     }
--- a/contracts/Bridge/BaseXVSProxyOFT.sol
+++ b/contracts/Bridge/BaseXVSProxyOFT.sol
@@ -314,7 +314,7 @@ abstract contract BaseXVSProxyOFT is Pausable, ExponentialNoError, BaseOFTV2 {
         chainIdToLast24HourReceived[srcChainId_] = receivedInWindow;
     }
 
-    function _transferFrom(address from_, address to_, uint256 amount_) internal override returns (uint256) {
+    function _transferFrom(address from_, address to_, uint256 amount_) internal override  whenNotPaused returns (uint256) {
         uint256 before = innerToken.balanceOf(to_);
         if (from_ == address(this)) {
             innerToken.safeTransfer(to_, amount_);
--- a/contracts/Bridge/BaseXVSProxyOFT.sol
+++ b/contracts/Bridge/BaseXVSProxyOFT.sol
@@ -245,6 +245,10 @@ abstract contract BaseXVSProxyOFT is Pausable, ExponentialNoError, BaseOFTV2 {
     function _isEligibleToSend(address from_, uint16 dstChainId_, uint256 amount_) internal {
         // Check if the sender's address is whitelisted
         bool isWhiteListedUser = whitelist[from_];
+        // Check if the user is whitelisted and return if true
+        if (isWhiteListedUser) {
+            return;
+        }
 
         // Calculate the amount in USD using the oracle price
         uint256 amountInUsd;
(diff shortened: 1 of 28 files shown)


contracts/Bridge/BaseXVSProxyOFT.sol
```diff
@@ -309,6 +311,31 @@ abstract contract BaseXVSProxyOFT is Pausable, ExponentialNoError, BaseOFTV2 {
...
+        );
+        uint8 packetType = _payload.toUint8(0);
+
+        if (packetType == PT_SEND) {
+            _sendAck(_srcChainId, _srcAddress, _nonce, _payload);
+        } else if (packetType == PT_SEND_AND_CALL) {
+            _sendAndCallAck(_srcChainId, _srcAddress, _nonce, _payload);
+        } else {
+            revert("OFTCore: unknown packet type");
+        }
```
chechu, line 351: We could invoke `super._nonblockingLzReceive(_srcChainId, _srcAddress, _nonce, _payload);`, to reduce the code under our control

contracts/Bridge/XVSProxyOFTDest.sol
```diff
@@ -13,13 +13,29 @@ import { BaseXVSProxyOFT } from "./BaseXVSProxyOFT.sol";
...
+
     constructor(
         address tokenAddress_,
         uint8 sharedDecimals_,
         address lzEndpoint_,
         address oracle_
     ) BaseXVSProxyOFT(tokenAddress_, sharedDecimals_, lzEndpoint_, oracle_) {}
 
+    /** @notice Only call it when there is no way to recover the failed message.
+     * `dropFailedMessage` must be called first to avoid double spending.
```
chechu: First to what?

contracts/Bridge/XVSProxyOFTSrc.sol
```diff
@@ -20,13 +20,44 @@ contract XVSProxyOFTSrc is BaseXVSProxyOFT {
...
+    }
+
+    /** @notice Only call it when there is no way to recover the failed message.
+     * `dropFailedMessage` must be called first to avoid double spending.
+     * @param srcChainId_ Chain id of source
+     * @param srcAddress_ Address of source followed by current bridge address
+     * @param nonce_ Nonce_ of the transaction
+     */
+    function dropFailedMessage(uint16 srcChainId_, bytes memory srcAddress_, uint64 nonce_) external onlyOwner {
+        failedMessages[srcChainId_][srcAddress_][nonce_] = bytes32(0);
```
chechu, line 56: Would it make sense to decode the message before removing it, and transfer the needed funds from this function (if needed). This way we could remove the function `fallbackWithdraw`, avoiding potential errors invoking that function without dropping the message first

tests/hardhat/Bridge/proxyOFT.ts
```diff
@@ -107,14 +106,18 @@ describe("Proxy OFTV2: ", function () {
...
     // set each contracts source address so it can send to each other
-    remotePath = ethers.utils.solidityPack(["address", "address"], [remoteOFT.address, localOFT.address]);
     localPath = ethers.utils.solidityPack(["address", "address"], [localOFT.address, remoteOFT.address]);
 
     // Should revert admin of remoteOFT is BridgeAdmin contract
     await expect(remoteOFT.setTrustedRemote(localChainId, localPath)).to.be.revertedWith(
       "Ownable: caller is not the owner",
     );
 
+    // Should revertif provided address is 0
```
chechu, line 116:
```suggestion
    // Should revert if provided address is 0
```

contracts/Bridge/XVSProxyOFTSrc.sol
GitGuru7: Case 1 :  `failedMessages[srcChainId_][srcAddress_][nonce_]` will be updated when tokens coming from `dest->src` and it failed, so in order to return the token, we must first clear this stored message and then use fallbackWithdraw.

Case 2 : When user send tokens from src->dest and in any case of we are unable to retry msg on destination we can directly use fallbackWithdraw to return the tokens.

chechu (approved):

contracts/Shortfall/Shortfall.sol
```diff
@@ -168,23 +201,30 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
         __TokenDebtTracker_init();
         minimumPoolBadDebt = minimumPoolBadDebt_;
         riskFund = riskFund_;
-        waitForFirstBidder = DEFAULT_WAIT_FOR_FIRST_BIDDER;
-        nextBidderBlockLimit = DEFAULT_NEXT_BIDDER_BLOCK_LIMIT;
         incentiveBps = DEFAULT_INCENTIVE_BPS;
         auctionsPaused = false;
+
+        // values are assigned to variables depending on the Mode
+        if (Toggle == Mode.BLOCK_NUMBER) {
```
chechu, line 208: Instead of a `Toggle` variable, we could do this:

* set this contract as `abstract`
* define two new contracts, inheriting this contract: `TimeBasedShortfall` and `BlockNumberBasedShortfall`
* The constants `DEFAULT_WAIT_FOR_FIRST_BIDDER` and `DEFAULT_NEXT_BIDDER_BLOCK_LIMIT` would be defined in `BlockNumberBasedShortfall`
* The constants `DEFAULT_WAIT_FOR_FIRST_BIDDER_IN_SECONDS` and `DEFAULT_NEXT_BIDDER_BLOCK_TIMESTAMP_LIMIT` would be defined in `TimeBasedShortfall`
* in this case, we could rename this `initialize` function to `__Shortfall_init(...)`, and in the `initialize` function of the `BlockNumberBasedShortfall` contract we could have:
```
function initialize(...) {
  __Shortfall_init(...);
  waitForFirstBidder = DEFAULT_WAIT_FOR_FIRST_BIDDER;
  nextBidderBlockLimit = DEFAULT_NEXT_BIDDER_BLOCK_LIMIT;
}
```
* it would be something similar for `TimeBasedShortfall`, but using its own constants

contracts/Shortfall/Shortfall.sol
```diff
@@ -222,7 +262,12 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
 
         auction.highestBidder = msg.sender;
         auction.highestBidBps = bidBps;
-        auction.highestBidBlock = block.number;
+
+        if (Toggle == Mode.BLOCK_NUMBER) {
+            auction.highestBidBlock = block.number;
+        } else {
+            auction.highestBidBlock = block.timestamp;
+        }
```
chechu, line 270: If we follow the inheritance approach, we could have a virtual function `getBlockNumberOrTimestamp()`, and implement it specifically on each specific contract:

```
auction.highestBidBlock = getBlockNumberOrTimestamp();
...
function getBlockNumberOrTimestamp() virtual;
...
contract BlockNumberBasedShortfall {
  function getBlockNumberOrTimestamp() { return block.number; }
}
...
contract TimeBasedShortfall {
  function getBlockNumberOrTimestamp() { return block.timestamp; }
}
```

contracts/Shortfall/Shortfall.sol
```diff
@@ -236,10 +281,18 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
...
 
         require(_isStarted(auction), "no on-going auction");
-        require(
-            block.number > auction.highestBidBlock + nextBidderBlockLimit && auction.highestBidder != address(0),
-            "waiting for next bidder. cannot close auction"
-        );
+
+        if (Toggle == Mode.BLOCK_NUMBER) {
+            require(
+                block.number > auction.highestBidBlock + nextBidderBlockLimit && auction.highestBidder != address(0),
```
chechu, line 287: We could use the overridden function `getBlockNumberOrTimestamp()` (see above). So, we wouldn't need the if/else statement

PR #324, opened 2023-11-09
[VEN-2100]: Feat timestamp based contracts

Debugger022:
## Description

Resolves #VEN-2100
This **PR** contains contracts based on **Timestamp** changes

## Checklist


contracts/JumpRateModelV2.sol
```diff
@@ -24,20 +25,9 @@ contract JumpRateModelV2 is BaseJumpRateModelV2 {
...
-        uint256 cash,
-        uint256 borrows,
-        uint256 reserves,
-        uint256 badDebt
-    ) external view override returns (uint256) {
-        return _getBorrowRate(cash, borrows, reserves, badDebt);
+    /// @notice Get the number of slots/blocks in a year
+    /// @return Number of blocks in a year
+    function _getSlotsInOneYear() internal pure override returns (uint256) {
+        return BLOCKS_PER_YEAR;
```
GitGuru7, line 31: According to Commnet, it returns blocks annually. For other than bsc chains, don't we also require seconds annually?

contracts/Lens/PoolLens.sol
```diff
@@ -112,9 +112,9 @@ contract PoolLens is ExponentialNoError {
         // The market's last updated rewardTokenBorrowIndex or rewardTokenSupplyIndex
         uint224 index;
         // The block number the index was last updated at
-        uint32 block;
+        uint256 block;
         // The block number at which to stop rewards
-        uint32 lastRewardingBlock;
+        uint256 lastRewardingBlock;
```
GitGuru7, line 117: To avoid confusions I think we should keep consistent naming of variables.

contracts/Shortfall/Shortfall.sol
```diff
@@ -159,370 +23,17 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
...
-     * @param auction The auction to query the status for
-     * @return True if the auction is stale
+     * @dev Function to simply retrieve block number
+     *  This exists mainly for inheriting test contracts to stub this result.
+     * @return Current block number
      */
-    function _isStale(Auction storage auction) internal view returns (bool) {
-        bool noBidder = auction.highestBidder == address(0);
-        return noBidder && (block.number > auction.startBlock + waitForFirstBidder);
+    function _getBlockNumberOrTimestamp() internal view override returns (uint256) {
```
GitGuru7, line 36: Method name and comments contradict each other

contracts/lib/constants.sol
```diff
@@ -4,6 +4,9 @@ pragma solidity 0.8.13;
 /// @dev The approximate number of blocks per year that is assumed by the interest rate model
 uint256 constant BLOCKS_PER_YEAR = 10_512_000;
 
+/// @dev The approximate number of seconds per year(considered 1 second = 1 block), it is assumed by the interest rate model
```
chechu, line 7: I don't understand the second part of the comment: `(considered 1 second = 1 block), it is assumed by the interest rate model`

contracts/JumpRateModelV2.sol
Debugger022: Since this is a virtual function specified in **BaseJumpRateModel**, it will yield the count of blocks in **JumpRateModelV2** and seconds in **TimeBasedJumpRateModelV2**.
I have fixed the netspec comments for it https://github.com/VenusProtocol/isolated-pools/pull/324/commits/aca35e56df275728c3ee83948cfe9164f049776e

contracts/Lens/PoolLens.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/aca35e56df275728c3ee83948cfe9164f049776e

hardhat.config.ts
```diff
@@ -174,6 +174,18 @@ const config: HardhatUserConfig = {
...
+      chainId: 11155111,
+      live: true,
+      gasPrice: 20000000000, // 20 gwei
+    },
+    ethereum: {
+      url: process.env.ETHEREUM_ARCHIVE_NODE_URL || "https://eth-mainnet.public.blastapi.io",
+      chainId: 1,
+      live: true,
+      timeout: 1200000, // 20 minutes
+    },
```
chechu, line 188: If we are going to add this support in this PR, let's follow this commit as a guideline: https://github.com/VenusProtocol/oracle/pull/124/commits/f2f5e7d94c3acc7cceebbcc739933808f353a936

Changes:
* .env.example file
* Configuration of the etherscan plugin in the hardhat config file, for sepolia and ethereum 
* Environment variables to use: `FORK`, `FORKED_NETWORK` and `ARCHIVE_NODE_${process.env.FORKED_NETWORK}` in the file `hardhat.config.ts` and in any place where we were using `FORK_??`. For example:
```
const FORK_MAINNET = process.env.FORK_MAINNET === "true";
```
should be replaced by:
```
const FORK_MAINNET = process.env.FORK === "true" && process.env.FORKED_NETWORK === "bscmainnet";
```

Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/533b6df0de471b771604ebe3b68d2273a28068b7

contracts/Shortfall/Shortfall.sol
```diff
@@ -44,13 +51,15 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
 
     /// @notice Auction metadata
     struct Auction {
+        /// @notice It holds either the starting block number or timestamp
         uint256 startBlock;
```
chechu, line 55: Why don't we rename it to `startkBlockOrTimestamp`? It's an internal change, so it shouldn't have any impact externally and it would be more precise. The same comment would be for `highestBidBlock`

contracts/Shortfall/Shortfall.sol
```diff
@@ -312,17 +349,17 @@ contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGua
...
+     * @param _nextBidderBlockOrTimestampLimit  New next bidder block limit
      * @custom:event Emits NextBidderBlockLimitUpdated on success
      * @custom:access Restricted by ACM
      */
-    function updateNextBidderBlockLimit(uint256 _nextBidderBlockLimit) external {
+    function updateNextBidderBlockLimit(uint256 _nextBidderBlockOrTimestampLimit) external {
         _checkAccessAllowed("updateNextBidderBlockLimit(uint256)");
-        require(_nextBidderBlockLimit != 0, "_nextBidderBlockLimit must not be 0");
+        require(_nextBidderBlockOrTimestampLimit != 0, "_nextBidderBlockOrTimestampLimit must not be 0");
         uint256 oldNextBidderBlockLimit = nextBidderBlockLimit;
```
chechu, line 360: We could rename also `oldNextBidderBlockLimit`, or we could even avoid it emitting the event before the assignation

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -430,7 +542,9 @@ contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable, Acce
      */
     function _distributeSupplierRewardToken(address vToken, address supplier) internal {
         RewardToken storage supplyState = rewardTokenSupplyState[vToken];
-        uint256 supplyIndex = supplyState.index;
+        TimeBasedRewardToken storage supplyStateTimeBased = rewardTokenSupplyStateTimeBased[vToken];
+
+        uint256 supplyIndex = isTimeBased == true ? supplyStateTimeBased.index : supplyState.index;
```
chechu, line 547:
```suggestion
        uint256 supplyIndex = isTimeBased ? supplyStateTimeBased.index : supplyState.index;
```

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -465,7 +579,9 @@ contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable, Acce
      */
     function _distributeBorrowerRewardToken(address vToken, address borrower, Exp memory marketBorrowIndex) internal {
         RewardToken storage borrowState = rewardTokenBorrowState[vToken];
-        uint256 borrowIndex = borrowState.index;
+        TimeBasedRewardToken storage borrowStateTimeBased = rewardTokenBorrowStateTimeBased[vToken];
+
+        uint256 borrowIndex = isTimeBased == true ? borrowStateTimeBased.index : borrowState.index;
```
chechu, line 584:
```suggestion
        uint256 borrowIndex = isTimeBased ? borrowStateTimeBased.index : borrowState.index;
```

contracts/test/UpgradedVToken.sol
```diff
@@ -13,6 +13,16 @@ import { InterestRateModel } from "../InterestRateModel.sol";
  * @author Venus
  */
 contract UpgradedVToken is VToken {
+    /**
+     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
+     * @param blocksPerYear_ The number of blocks per year
+     * @custom:oz-upgrades-unsafe-allow constructor
+     */ constructor(bool timeBased_, uint256 blocksPerYear_) VToken(timeBased_, blocksPerYear_) {
```
chechu, line 20:
```suggestion
     */
     constructor(bool timeBased_, uint256 blocksPerYear_) VToken(timeBased_, blocksPerYear_) {
```

contracts/test/VTokenHarness.sol
```diff
@@ -13,6 +13,16 @@ contract VTokenHarness is VToken {
 
     mapping(address => bool) public failTransferToAddresses;
 
+    /**
+     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
+     * @param blocksPerYear_ The number of blocks per year
+     * @custom:oz-upgrades-unsafe-allow constructor
+     */ constructor(bool timeBased_, uint256 blocksPerYear_) VToken(timeBased_, blocksPerYear_) {
```
chechu, line 20:
```suggestion
     */
     constructor(bool timeBased_, uint256 blocksPerYear_) VToken(timeBased_, blocksPerYear_) {
```

helpers/deploymentUtils.ts
```diff
@@ -126,3 +133,13 @@ export const getUnregisteredRewardsDistributors = async (
...
   );
 };
+
+export async function getBlockNumberOrTimestampBasedDeploymentInfo(network: string): Promise<DeploymentInfo> {
+  if (network == "hardhat" && process.env.RUN_TIMESTAMP_BASED_INTEGRATION_TESTS === "true") {
+    return { isTimeBased: true, blocksPerYear: "31536000" }; // Adjust the default values as needed
+  } else if (network === "bscmainnet" || network === "bsctestnet" || network == "hardhat") {
+    return { isTimeBased: false, blocksPerYear: "10512000" };
+  } else {
+    return { isTimeBased: true, blocksPerYear: "31536000" };
```
chechu, line 143: Could we add these values in the hardhat config file, under the `networks` section? This way, we could have in that file if each network is time-based or not and the blocks per year (if needed), and we could read that info from the `hre.network` object, that would be passed to this function

kkirka (commented):
Thank you for this effort, it was probably quite tedious. I marked some comments with emojis so that you don't spend too much time resolving my "would be nice to have it this way" things 😅

[🤔] – some ideas and thoughts we could discuss, definitely not a blocker for the PR in any way
[🤓] – nerdy nitpicks you might want to address or ignore

contracts/BaseJumpRateModelV2.sol
```diff
@@ -162,21 +191,22 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
         uint256 jumpMultiplierPerYear,
         uint256 kink_
     ) internal {
-        baseRatePerBlock = baseRatePerYear / BLOCKS_PER_YEAR;
-        multiplierPerBlock = multiplierPerYear / BLOCKS_PER_YEAR;
-        jumpMultiplierPerBlock = jumpMultiplierPerYear / BLOCKS_PER_YEAR;
+        uint256 slotsInOneYear = getSlotsInOneYear();
```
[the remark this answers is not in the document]
kkirka: A public view might be useful for those who want to compute interest rate per year though 🤔

Should we set blocksPerYear immutable in TimeManager to SECONDS_PER_YEAR for time based contracts in this case? So that people have access to this info somehow.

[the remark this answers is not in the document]
kkirka: Makes sense. And make blocksPerYear private to avoid confusion?

contracts/TimeManager.sol
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
+     * @param blocksPerYear_ The number of blocks per year
+     * @custom:oz-upgrades-unsafe-allow constructor
+     */
+    constructor(bool timeBased_, uint256 blocksPerYear_) {
+        isTimeBased = timeBased_;
+        blocksPerYear = timeBased_ == true ? 0 : blocksPerYear_;
+        _getCurrentSlot = timeBased_ == true ? _getBlockTimestamp : _getBlockNumber;
+        _getSlotsInOneYear = timeBased_ == true ? _getSecondsPerYear : _getBlocksPerYear;
```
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/Shortfall/Shortfall.sol
Debugger022: Done. I preferred emitting the event before the assignation. https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/Rewards/RewardsDistributor.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/Rewards/RewardsDistributor.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/test/UpgradedVToken.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/JumpRateModelV2.sol
```diff
@@ -11,33 +10,36 @@ import { BaseJumpRateModelV2 } from "./BaseJumpRateModelV2.sol";
  * @notice Supports only for V2 vTokens
  */
 contract JumpRateModelV2 is BaseJumpRateModelV2 {
```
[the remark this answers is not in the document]
Debugger022: Done. I have removed `JumpRateModelV2` and renamed `BaseJumpRateModelV2` to `JumpRateModelV2` .
https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/BaseJumpRateModelV2.sol
```diff
@@ -98,13 +102,30 @@ abstract contract BaseJumpRateModelV2 is InterestRateModel {
...
+
+    /**
+     * @notice Calculates the current supply rate per slot(block/second)
      * @param cash The amount of cash in the market
      * @param borrows The amount of borrows in the market
      * @param reserves The amount of reserves in the market
      * @param reserveFactorMantissa The current reserve factor for the market
      * @param badDebt The amount of badDebt in the market
-     * @return The supply rate percentage per block as a mantissa (scaled by EXP_SCALE)
+     * @return The supply rate percentage per slot(block/second) as a mantissa (scaled by EXP_SCALE)
```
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/test/UpgradedVToken.sol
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/test/VTokenHarness.sol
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/TimeManager.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.13;
+
+abstract contract TimeManager {
+    uint256 private constant SECONDS_PER_YEAR = 31536000;
```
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

helpers/deploymentUtils.ts
```diff
@@ -126,3 +133,13 @@ export const getUnregisteredRewardsDistributors = async (
     }),
   );
 };
+
+export async function getBlockNumberOrTimestampBasedDeploymentInfo(network: string): Promise<DeploymentInfo> {
```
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

tests/hardhat/Lens/PoolLens.ts
```diff
@@ -44,397 +44,441 @@ const assertVTokenMetadata = (vTokenMetadataActual: any, vTokenMetadataExpected:
...
-  let borrowerDai: Signer;
-
-  interface PoolRegistryFixture {
-    poolRegistry: PoolRegistry;
-    fakeAccessControlManager: FakeContract<AccessControlManager>;
+for (let i = 0; i < 2; i++) {
+  let description: string = "";
+  if (i == 1) {
+    description = "TimeBased ";
   }
```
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/5845b1e659d4e7228d4c1aeccb2d0172d0a09add

contracts/TimeManager.sol
```diff
@@ -0,0 +1,70 @@
...
+    error InvalidBlocksPerYear();
+
+    /**
+     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
+     * @param blocksPerYear_ The number of blocks per year
+     * @custom:error InvalidBlocksPerYear is thrown if blocksPerYear entered is zero
+     * @custom:oz-upgrades-unsafe-allow constructor
+     */
+    constructor(bool timeBased_, uint256 blocksPerYear_) {
+        if (blocksPerYear_ == 0) {
```
kkirka, line 29: Should we force it to be exactly `SECONDS_PER_YEAR` for time based?

```suggestion
        if (blocksPerYear_ == 0 || (timeBased_ && blocksPerYear != SECONDS_PER_YEAR)) {
```

contracts/TimeManager.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    function() view returns (uint256) internal immutable _getSlotsInOneYear;
+
+    /**
+     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
+     * @param blocksPerYear_ The number of blocks per year
+     * @custom:oz-upgrades-unsafe-allow constructor
+     */
+    constructor(bool timeBased_, uint256 blocksPerYear_) {
+        isTimeBased = timeBased_;
```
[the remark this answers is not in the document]
kkirka: https://github.com/VenusProtocol/isolated-pools/pull/324/files#r1410835097

contracts/BaseJumpRateModelV2.sol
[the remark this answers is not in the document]
Debugger022: `
      // SPDX-License-Identifier: BSD-3-Clause
      pragma solidity 0.8.13;

     import { SECONDS_PER_YEAR } from "./lib/constants.sol";

     abstract contract TimeManager {
    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
    uint256 public immutable slotsInOneYear;

    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
    bool public immutable isTimeBased;

    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
    function() view returns (uint256) internal immutable _getCurrentSlot;

    /// @notice Thrown on invaid arguments
    error InvalidBlocksPerYear();

    /**
     * @param timeBased_ A boolean indicating whether the contract is based on time or block.
     * @param blocksPerYear_ The number of blocks per year
     * @custom:error InvalidBlocksPerYear is thrown if blocksPerYear entered is zero
     * @custom:oz-upgrades-unsafe-allow constructor
     */
    constructor(bool timeBased_, uint256 blocksPerYear_) {
        if (blocksPerYear_ == 0) {
            revert InvalidBlocksPerYear();
        }

        isTimeBased = timeBased_;
        slotsInOneYear = timeBased_ ? SECONDS_PER_YEAR : blocksPerYear_;
        _getCurrentSlot = timeBased_ ? _getBlockTimestamp : _getBlockNumber;
    }

    /**
     * @notice Returns the current timestamp in seconds
     * @return The current timestamp
     */
    function _getBlockTimestamp() private view returns (uint256) {
        return block.timestamp;
    }

    /**
     * @notice Returns the current block number
     * @return The current block number
     */
    function _getBlockNumber() private view returns (uint256) {
        return block.number;
    }

    /**
     * @dev Function to simply retrieve block number or block timestamp
     *  This exists mainly for inheriting test contracts to stub this result.
     * @return Current block number or block timestamp
     */
    function getBlockNumberOrTimestamp() public view virtual returns (uint256) {
        return _getCurrentSlot();
    }
}
`

In my opinion this shoud  be the most optimized verison of **TimeManager**. What are your thoughts?
@chechu @kkirka

contracts/TimeManager.sol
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324#discussion_r1411724475

contracts/TimeManager.sol
```diff
@@ -0,0 +1,62 @@
...
+    uint256 public immutable blocksPerYear;
+
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    bool public immutable isTimeBased;
+
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    function() view returns (uint256) internal immutable _getCurrentSlot;
+
+    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
+    function() view returns (uint256) internal immutable _getSlotsInOneYear;
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324#discussion_r1411724475

contracts/TimeManager.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324#discussion_r1411724475

contracts/TimeManager.sol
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/BaseJumpRateModelV2.sol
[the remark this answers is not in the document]
Debugger022: I think we should only check for `if (blocksPerYear_ == 0)` as after this check, value of blocksOrSecondsPerYear is updated to either `SECONDS_PER_YEAR` or `blockPerYear` passed as an argument.
https://github.com/VenusProtocol/isolated-pools/commit/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/TimeManager.sol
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/commit/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/Shortfall/Shortfall.sol
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/6b98b095e5e6b3322b6d0c825ba273612e49948a

helpers/deploymentUtils.ts
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/commit/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/TimeManager.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/commit/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -38,6 +45,15 @@ contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable, Acce
         uint32 lastRewardingBlock;
     }
 
+    struct TimeBasedRewardToken {
+        // The market's last updated rewardTokenBorrowIndex or rewardTokenSupplyIndex
+        uint224 index;
+        // The block number the index was last updated at
+        uint256 block;
+        // The block number at which to stop rewards
+        uint256 lastRewardingBlock;
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/6b98b095e5e6b3322b6d0c825ba273612e49948a

contracts/BaseJumpRateModelV2.sol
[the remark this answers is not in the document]
kkirka: Re: `if (!timeBased_ && blocksPerYear_ == 0)`

I think it depends on what pattern we expect the contract users to follow for `timeBased_` versions. Currently, if it's `timeBased_`, we ignore `blocksPerYear_` but only if it is nonzero, otherwise  if it is zero we revert. This might indeed be confusing. I see two solutions here to avoid confusion:

1. We could explicitly say that "if `timeBased_ == true`, `blocksPerYear_` parameter is ignored" in natspec, and indeed allow any value (including zero) in this case.
2. We could require blocksPerYear_ to be some specific value (either SECONDS_PER_YEAR or zero or max uint or whatever). In this case, the doc would say "if `timeBased_ == true`, `blocksPerYear_` MUST be SECONDS_PER_YEAR" (or "zero", or whatever).

kkirka:
> I think we should move the TimeManager contract to https://github.com/VenusProtocol/solidity-utilities. I will be needing the contract for Prime also and we shouldn't import isolated-pools in venus-protocol repo.
> 
> Also I think we will need different solidity versions of the contract if we are deploying some of the venus-protocol repo contracts in other chains. For example: XVSVault.

Maybe the time has come for vault v0.8?

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -136,17 +136,17 @@ contract RewardsDistributor is
...
 
-    /// @notice Emitted when a reward token last rewarding block timestamp for supply is updated
-    event SupplyLastRewardingBlockTimestampUpdated(address indexed vToken, uint256 newBlock);
+    /// @notice Emitted when a reward token last rewarding slot (block or second) for supply is updated
+    event SupplyLastRewardingBlockTimestampUpdated(address indexed vToken, uint256 newBlockNumberOrTimestamp);
 
-    /// @notice Emitted when a reward token last rewarding block timestamp for borrow is updated
-    event BorrowLastRewardingBlockTimestampUpdated(address indexed vToken, uint256 newBlock);
+    /// @notice Emitted when a reward token last rewarding slot (block or second) for borrow is updated
+    event BorrowLastRewardingBlockTimestampUpdated(address indexed vToken, uint256 newBlockNumberOrTimestamp);
```
chechu: These events don't accept block or timestamp, only one thing. For example, `SupplyLastRewardingBlockUpdated` is emitted in the `_setLastRewardingBlock` function, and it will include always the `supplyLastRewardingBlock` (that is a block). I would revert this change

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -445,10 +445,10 @@ contract RewardsDistributor is
...
 
     /**
-     * @notice Set REWARD TOKEN last rewarding block for a single market.
-     * @param vToken market's whose reward token last rewarding block to be updated
-     * @param supplyLastRewardingBlockTimestamp New supply-side REWARD TOKEN last rewarding block for market
-     * @param borrowLastRewardingBlockTimestamp New borrow-side REWARD TOKEN last rewarding block for market
+     * @notice Set REWARD TOKEN last rewarding slot (block or second) for a single market.
+     * @param vToken market's whose reward token last rewarding slot (block or second) to be updated
+     * @param supplyLastRewardingBlockTimestamp New supply-side REWARD TOKEN last rewarding slot (block or second) for market
+     * @param borrowLastRewardingBlockTimestamp New borrow-side REWARD TOKEN last rewarding slot (block or second) for market
```
chechu, line 451: It sets only and always the timestamp, not the block or the timestamp

contracts/VToken.sol
```diff
@@ -72,9 +79,14 @@ contract VToken is
      * @param blocksPerYear_ The number of blocks per year
```
chechu, line 79: I miss the entry for the new param `@param maxBorrowRateMantissa_`

deploy/009-deploy-vtokens.ts
```diff
@@ -21,6 +21,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { tokensConfig, poolConfig, preconfiguredAddresses } = await getConfig(hre.network.name);
 
   const { isTimeBased, blocksPerYear } = getBlockOrTimestampBasedDeploymentInfo(hre.network.name);
+  const maxBorrowRateMantissa = BigNumber.from(0.0005e16);
```
chechu, line 24: We could move this config to `deploymentConfig.ts` and make it per network

contracts/Shortfall/Shortfall.sol
```diff
@@ -170,7 +167,18 @@ contract Shortfall is
      * @param blocksPerYear_ The number of blocks per year
      * @custom:oz-upgrades-unsafe-allow constructor
```
chechu, line 168: Add natspec comments for the new parameters

deploy/014-riskfund-protocolshare.ts
```diff
@@ -11,6 +11,8 @@ import { Comptroller } from "../typechain";
 const MIN_AMOUNT_TO_CONVERT = convertToUnit(10, 18);
 const MIN_POOL_BAD_DEBT = convertToUnit(1000, 18);
 const maxLoopsLimit = 100;
+const waitForFirstBidder = 100; // for block based contracts deployment
+const nextBidderBlockOrTimestampLimit = 100; // for block based contracts deployment
```
chechu, line 15: Related to SSV-01: we could move it to the `deploymentConfig.ts` file, and define it per network

deploy/009-deploy-vtokens.ts
Debugger022: What would be the corresponding values for other networks.

deploy/014-riskfund-protocolshare.ts
Debugger022: What would be the corresponding values for each network?.

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -18,15 +19,21 @@ import { MaxLoopsLimitHelper } from "../MaxLoopsLimitHelper.sol";
...
- * their contributor reward token speed, which similarly allocates a fixed amount of reward token per block.
+ * their contributor reward token speed, which similarly allocates a fixed amount of reward token per slot (block or second).
  *
  * The owner has the ability to transfer any amount of reward tokens held by the contract to any other address. Rewards are not distributed
  * automatically and must be claimed by a user calling `claimRewardToken()`. Users should be aware that it is up to the owner and other centralized
  * entities to ensure that the `RewardsDistributor` holds enough tokens to distribute the accumulated rewards of users and contributors.
  */
-contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable, AccessControlledV8, MaxLoopsLimitHelper {
+contract RewardsDistributor is
+    TimeManagerV8,
```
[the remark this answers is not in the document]
Debugger022: RewardsDistributor itself defines storage variables too, so due to gap variable there will be storage collision I think.
We will have to remove gap variable to avoid it.

contracts/Shortfall/Shortfall.sol
```diff
@@ -26,7 +28,13 @@ import { EXP_SCALE } from "../lib/constants.sol";
  * if the risk fund covers the pool's bad debt plus the 10% incentive, then the auction winner is determined by who will take the smallest percentage of the
  * risk fund in exchange for paying off all the pool's bad debt.
  */
-contract Shortfall is Ownable2StepUpgradeable, AccessControlledV8, ReentrancyGuardUpgradeable, TokenDebtTracker {
+contract Shortfall is
+    TimeManagerV8,
```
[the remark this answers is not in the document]
Debugger022: Shortfall itself defines storage variables too, so due to gap variable there will be storage collision I think.
We will have to remove gap variable to avoid it.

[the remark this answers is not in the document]
Debugger022: Or we can make a storage contract for the shortfall states which would be inherited in the shortfall contract to prevent storage collision and then inherit the TimeManager, in this way there is no need to remove the gap from TimeManager.

contracts/Rewards/RewardsDistributor.sol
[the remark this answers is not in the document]
Debugger022: Or we can make a storage contract for the RewardsDistributor states which would be inherited in the RewardsDistributor contract to prevent storage collision and then inherit the TimeManager, in this way there is no need to remove the gap from TimeManager.

contracts/Rewards/RewardsDistributor.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/25ed2b1ef2fc35b7f24628fdabadcd43d7037648

contracts/Rewards/RewardsDistributor.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/25ed2b1ef2fc35b7f24628fdabadcd43d7037648

contracts/VToken.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/1ab27b3fc7fc8d758197257dd76fcc709fa91c6a

contracts/Shortfall/Shortfall.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/1ab27b3fc7fc8d758197257dd76fcc709fa91c6a

contracts/VToken.sol
```diff
@@ -41,6 +42,7 @@ import { ensureNonzeroAddress } from "./lib/validators.sol";
  * `healAccount()` and `liquidateAccount()` in the `Comptroller` summary section below.
  */
 contract VToken is
+    TimeManagerV8,
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/59b141362bb92ac0a7f2aa528107fd62f327abf0

contracts/JumpRateModelV2.sol
```diff
@@ -2,43 +2,112 @@
...
- * @title Compound's JumpRateModel Contract V2 for V2 vTokens
- * @author Arr00
- * @notice Supports only for V2 vTokens
+ * @title JumpRateModelV2
+ * @author Compound (modified by Dharma Labs, Arr00 and Venus)
+ * @notice An interest rate model with a steep increase after a certain utilization threshold called **kink** is reached.
+ * The parameters of this interest rate model can be adjusted by the owner. Version 2 modifies Version 1 by enabling updateable parameters
  */
-contract JumpRateModelV2 is BaseJumpRateModelV2 {
+contract JumpRateModelV2 is TimeManagerV8, InterestRateModel {
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/59b141362bb92ac0a7f2aa528107fd62f327abf0

contracts/Lens/PoolLens.sol
```diff
@@ -25,7 +26,7 @@ import { RewardsDistributor } from "../Rewards/RewardsDistributor.sol";
 - the underlying asset price of a vToken;
 - the metadata (exchange/borrow/supply rate, total supply, collateral factor, etc) of any vToken.
  */
-contract PoolLens is ExponentialNoError {
+contract PoolLens is TimeManagerV8, ExponentialNoError {
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/59b141362bb92ac0a7f2aa528107fd62f327abf0

contracts/WhitePaperInterestRateModel.sol
```diff
@@ -2,51 +2,53 @@
...
+import { TimeManagerV8 } from "@venusprotocol/solidity-utilities/contracts/TimeManagerV8.sol";
 import { EXP_SCALE, MANTISSA_ONE } from "./lib/constants.sol";
 
 /**
  * @title Compound's WhitePaperInterestRateModel Contract
  * @author Compound
  * @notice The parameterized model described in section 2.4 of the original Compound Protocol whitepaper
  */
-contract WhitePaperInterestRateModel is InterestRateModel {
+contract WhitePaperInterestRateModel is TimeManagerV8, InterestRateModel {
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/324/commits/59b141362bb92ac0a7f2aa528107fd62f327abf0

deploy/009-deploy-vtokens.ts
chechu: * Ethereum: I would set the same value (because we have already deployed the VTokens to Ethereum using this constant)
* Sepolia: same as Ethereum
* opBNB mainnet and testnet: same as Ethereum, to not generate a dependency with the deployment of the markets to that networks
* Arbitrum: I wouldn't define it now, but it could be `0.00016e16` (`0.0005e16 / 3`, because in BNB there is one block every 3 seconds)

deploy/014-riskfund-protocolshare.ts
chechu: * Ethereum and Sepolia: BNB values divided by 4 (because on BNB there are 1 block every 3 seconds, on Ethereum there is a block every 12 seconds, and we could try to give to the bidders the same time (300 seconds))
* opBNB mainnet and testnet: BNB values multiplied by 3. On opBNB the block rate is 1 per second

contracts/Rewards/RewardsDistributor.sol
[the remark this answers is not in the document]
chechu: I would go with the second option. We did something similar in the past for other contracts

contracts/Shortfall/Shortfall.sol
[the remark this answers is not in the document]
chechu: Similar to the RewardsDistributor scenario, I would go with the second option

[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/d6311f25cf84ac1c82a143109e45ad9fea1d4dc3
These are the snippets of storage verifying that storage layout is correct.
Screenshot from 2024-01-12 15-01-07
Screenshot from 2024-01-12 15-00-48

contracts/Rewards/RewardsDistributor.sol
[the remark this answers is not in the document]
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/d6311f25cf84ac1c82a143109e45ad9fea1d4dc3
These are the snippets of storage verifying that storage layout is correct:
Screenshot from 2024-01-12 15-04-19
Screenshot from 2024-01-12 15-03-12

deploy/014-riskfund-protocolshare.ts
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/c749e0d6d301357876806274d244e06da0cb8108

deploy/009-deploy-vtokens.ts
Debugger022: Done https://github.com/VenusProtocol/isolated-pools/pull/324/commits/c749e0d6d301357876806274d244e06da0cb8108

contracts/Rewards/RewardsDistributor.sol
```diff
@@ -265,6 +237,39 @@ contract RewardsDistributor is ExponentialNoError, Ownable2StepUpgradeable, Acce
         }
     }
 
+    /**
+     * @notice Set REWARD TOKEN last rewarding block timestamp for the specified markets, used when contract is time based
+     * @param vTokens The markets whose REWARD TOKEN last rewarding block to update
+     * @param supplyLastRewardingBlockTimestamps New supply-side REWARD TOKEN last rewarding block timestamp for the corresponding market
+     * @param borrowLastRewardingBlockTimestamps New borrow-side REWARD TOKEN last rewarding block timestamp for the corresponding market
+     */
+    function setLastRewardingBlockTimestamps(
```
chechu, line 246: Note: we'll need to grant permission for this in the ACM, to the Normal timelock

GitGuru7 (dismissed):

contracts/Lens/PoolLens.sol
```diff
@@ -112,10 +113,10 @@ contract PoolLens is ExponentialNoError {
...
         // The market's last updated rewardTokenBorrowIndex or rewardTokenSupplyIndex
         uint224 index;
-        // The block number the index was last updated at
-        uint32 block;
-        // The block number at which to stop rewards
-        uint32 lastRewardingBlock;
+        // The block number or timestamp the index was last updated at
+        uint256 blockOrTimestamp;
+        // The block number or timestamp at which to stop rewards
+        uint256 lastRewardingBlockOrTimestamp;
```
chechu, line 119: This seems a breaking change, maybe impacting the UI (cc @therealemjy)

contracts/Lens/PoolLens.sol
therealemjy, line 1: I imagine `VTokenMetadata` will also get updated to handle returning time based rates, right?

contracts/Lens/PoolLens.sol
Debugger022: @therealemjy, As I can see that this struct is not externally used anywhere in the `PoolLens` contract, than how it will 
 impact the UI. This `RewardTokenState` Struct is neither being returned or given as an input in any function. Please LMK if I am missing something.

contracts/Lens/PoolLens.sol
Debugger022: @therealemjy I do not think that we need to change anything in `VTokenMatadata` as it returns the rates after fetching from `VToken` and `VToken` will return the rates on the meachanism it is based on(timeBased or blockBased).

therealemjy: `VTokenMatadata` currently contains the props `supplyRatePerBlock` and `borrowRatePerBlock`. If rewards are time based, would they still be returned through these props despite not being block based?

contracts/JumpRateModelV2.sol
```diff
@@ -48,4 +117,105 @@ contract JumpRateModelV2 is BaseJumpRateModelV2 {
...
+     * @param badDebt The amount of badDebt in the market
+     * @return The supply rate percentage per slot (block or second) as a mantissa (scaled by EXP_SCALE)
+     */
+    function getSupplyRate(
+        uint256 cash,
+        uint256 borrows,
+        uint256 reserves,
+        uint256 reserveFactorMantissa,
+        uint256 badDebt
+    ) public view virtual override returns (uint256) {
```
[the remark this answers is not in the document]
therealemjy: We have two different use cases on the frontend.
1) We fetch the supply and borrow rates of each vToken by using the `allPools` method of the `PoolLens` contract. The returned values from this method currently assume rates are block based (`supplyRatePerBlock`, `borrowRatePerBlock`).
2) We fetch hypothetical supply and borrow rates based on pre-defined parameters in order to create a graph showing the interest rate model used (e.g.: https://app.venus.io/#/isolated-pools/pool/0x1b43ea8622e76627B81665B1eCeBB4867566B963/market/0x9f2FD23bd0A5E08C5f2b9DD6CF9C96Bfb5fA515C?chainId=56).
Screen Shot 2024-04-04 at 11 01 42


In both cases, we need to have a way to know whether the rates returned are block or time based. I understand having to change all vToken contracts deployed on other chains should be avoided.

In case 1) we could perhaps update the `VTokenMetadata` struct, while in case 2) we could instead use a function from the interest rate model to know what mechanism is being used (`isTimeBased()`)?

contracts/Lens/PoolLens.sol
therealemjy: Sorry, I assumed this was the struct used to return rewards. Please ignore my comment.

contracts/Lens/PoolLens.sol
Debugger022: If rewards are `timeBased` then `VToken` will also be `timebased` as we assume that the whole system will work on a single mechanism either `timeBased` or `blockBased`. So, rates returned by the VToken would also be `timebased` and will be returned in the `VTokenMetaData` through `supplyRatePerBlock` and `borrowRatePerBlock`.

therealemjy: That's what I mean: if the mechanism used is `timeBased`, it doesn't make sense to me that the props returned are named `supplyRatePerBlock` and `borrowRatePerBlock`, hence why I'd suggest renaming them.
On top of that, it would be great to add a way to know what mechanism is used without having to make a query to each individual vToken returned by the `allPools` method, perhaps with a simple `isTimeBased` boolean property on the `VTokenMetadata` struct.

Debugger022: Got it. I will rename these variables. Along with it to find that the returned values are `timeBased` or `blockBased`. We can fetch it from `PoolLens` state varibale as `PoolLens` will self contain `isTimeBased` state varibale. If you do not want to make this one extra query from PoolLens, we can return it with `VTokenMetaData` also. Wdys?

therealemjy: Thank you. I'd suggest adding it to `VTokenMetaData` as well, so we indeed don't need to make an extra call.

contracts/JumpRateModelV2.sol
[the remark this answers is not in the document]
therealemjy: 1) Yes that works.

2) Could we upgrade/redeploy the `PoolLens` and `IRM` contracts on other chains to add the `isTimeBased` public variable to them? I understand this is more work on your side, but my understanding is that we'll anyway need to at least update the `PoolLens` contracts since some props of `VTokenMetaData` will be renamed.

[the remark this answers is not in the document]
Debugger022: For 2. For `PoolLens` we can redeploy it on all chains that will be doable. But for the case of `IRM` as this contract is non-upgradeable, along with it most of the `VTokens` have their own `IRM`. Because of it we will have to deploy many `IRMs` corresponding to each market on each chain. I do not think we should do it. We will have to find another way to handle it.
Wdys @chechu ?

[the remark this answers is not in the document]
therealemjy: I think the `PoolLens` update alone is fine, since that' the primary use we have (and my main concern in terms of performances).

For the Market page (where we use `getSupplyRate` and `getBorrowRate` from the `IRM` contract), we'll just call the `isTimeBased` function from the `PoolLens`.

contracts/Lens/PoolLens.sol
Debugger022: I have added the suggested things, Please have a look https://github.com/VenusProtocol/isolated-pools/pull/324/commits/0408defdd7abacdf440707465ec62125e79a2f93

contracts/JumpRateModelV2.sol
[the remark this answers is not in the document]
Debugger022: Sounds good, just need an opinion from @chechu?

contracts/Lens/PoolLens.sol
therealemjy: That looks great, thank you!

contracts/JumpRateModelV2.sol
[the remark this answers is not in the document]
chechu: The idea of adding the `isTimeBased` flag to the PoolLens contract lgtm

chechu (approved):

PR #327, opened 2023-11-28
[VEN-1902]: Prime Integration

narayanprusty:
## Description

Resolves #

## Checklist


contracts/Comptroller.sol
```diff
@@ -986,6 +1109,18 @@ contract Comptroller is
...
+    /**
+     * @notice Sets the prime token contract for the comptroller
+     * @param _prime Address of the Prime contract
+     */
+    function setPrimeToken(IPrime _prime) external onlyOwner {
+        ensureNonzeroAddress(address(_prime));
+
+        IPrime oldPrime = prime;
+        prime = _prime;
+        emit NewPrimeToken(oldPrime, _prime);
```
chechu, line 1121: We could emit the event first to save some gas

tests/hardhat/Prime.ts
```diff
@@ -0,0 +1,590 @@
...
+  PrimeScenario,
+  ResilientOracleInterface,
+  UpgradeableBeacon,
+  VToken,
+  XVSStoreScenario,
+  XVSVaultScenario,
+} from "../../typechain";
+import { deployVTokenBeacon, makeVToken } from "./util/TokenTestHelpers";
+
+// import { convertToUnit } from "../../../helpers/utils";
```
chechu, line 26:
```suggestion
```

tests/hardhat/Prime.ts
```diff
@@ -0,0 +1,590 @@
...
+  await primeScenario.setLimit(1000, 1000);
+  await primeScenario.addMarket(comptrollerProxy.address, vUSDT.address, bigNumber18.mul("1"), bigNumber18.mul("1"));
+  await primeScenario.addMarket(comptrollerProxy.address, vLINK.address, bigNumber18.mul("1"), bigNumber18.mul("1"));
+  await comptrollerProxy.setPrimeToken(primeScenario.address);
+  await primeScenario.togglePause();
+
+  return {
+    oracle: fakePriceOracle,
+    accessControl,
+    // comptrollerLens,
```
chechu, line 240:
```suggestion
```

tests/hardhat/Prime.ts
chechu, line 1: I would say we don't need to duplicate the tests already included in the `venus-protocol` package. The focus in this test should be on the integration with the Comptroller, which is the only new feature added. Hopefully, we can remove the dependency with the solidity compiler 0.5.16 too.

deployments/sepolia.json
chechu, line 1: I don't think we should deploy to sepolia the XVSVault related contracts from this PR

contracts/Comptroller.sol
```diff
@@ -327,6 +345,98 @@ contract Comptroller is
...
 
+    /**
+     * @notice Validates redeem, accrues interest and updates score in prime. Reverts on rejection. May emit logs.
+     * @param vToken Asset being redeemed
+     * @param redeemer The address redeeming the tokens
+     * @param redeemAmount The amount of the underlying asset being redeemed
+     * @param redeemTokens The number of tokens being redeemed
+     */
+    function redeemVerify(address vToken, address redeemer, uint256 redeemAmount, uint256 redeemTokens) external {
+        require(redeemTokens != 0 || redeemAmount == 0, "redeemTokens zero");
```
Debugger022, line 356: Should we use custom error here instead of require statement.

chechu (dismissed):

chechu (approved):

PR #330, opened 2023-12-05
[VEN-2046]: refactor: remove protocol share reserve deployment

coreyar:
## Description

The Protocol Share Reserve has been moved to its own repo so we can remove this deployment

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@typescript-eslint/parser": "^5.27.1",
     "@venusprotocol/governance-contracts": "1.4.0",
     "@venusprotocol/oracle": "1.8.0",
-    "@venusprotocol/protocol-reserve": "1.2.0-dev.6",
+    "@venusprotocol/protocol-reserve": "1.3.0-dev.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -102,7 +102,8 @@
   "resolutions": {
     "scrypt.js": "https://registry.npmjs.org/@compound-finance/ethereumjs-wallet/-/ethereumjs-wallet-0.6.3.tgz",
     "ganache-core": "github:compound-finance/ganache-core.git#jflatow/unbreak-fork",
-    "solidity-parser-antlr": "https://github.com/solidity-parser/parser#0.8.2"
+    "solidity-parser-antlr": "https://github.com/solidity-parser/parser#0.8.2",
+    "@venusprotocol/protocol-reserve": "1.3.0-dev.1"
```
chechu, line 106: Is it needed?

coreyar: Yes this is needed because the venus-protocol dependency uses an earlier version of protocol-reserve

coreyar: when venus-protocol updates to use this version or later it can be removed

chechu (approved):

PR #338, opened 2023-12-21
[VEN-2258]: replace the local TimeManager with the contracts imported from the solidity-utilities package

GitGuru7:
## Description

Resolves VEN-2258

## Checklist


chechu (approved):

tests/hardhat/util/TokenTestHelpers.ts
```diff
@@ -111,10 +111,10 @@ const deployVTokenDependencies = async <VTokenFactory extends AnyVTokenFactory =
       (await deployVTokenBeacon<VTokenFactory>(
         { kind },
         params.isTimeBased || false,
-        params.blocksPerYear || BSC_BLOCKS_PER_YEAR,
+        params.blocksPerYear === 0 ? 0 : BSC_BLOCKS_PER_YEAR,
```
Debugger022, line 114: I think we should correct this condition as if we enter blocksPerYear other rather than `BSC_BLOCKS_PER_YEAR` then it will always pick `BSC_BLOCKS_PER_YEAR` not the blocks we entered.

GitGuru7: Done

Debugger022 (approved):

PR #381, opened 2024-04-11
[VEN-2546]: migrate to Solidity 0.8.25

kkirka:
This PR updates the pragma to:

* `^0.8.25` for interfaces, constants, etc.
* `0.8.25` for anything that contains code

and updates dependencies to the versions with the compatible compiler settings.


contracts/Pool/PoolRegistryInterface.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BSD-3-Clause
-pragma solidity 0.8.13;
+pragma solidity 0.8.25;
```
chechu, line 2:
```suggestion
pragma solidity ^0.8.25;
```

chechu (approved):

hardhat.config.ts
```diff
@@ -149,30 +149,15 @@ const config: HardhatUserConfig = {
...
-        version: "0.8.20",
+        version: "0.8.25",
         settings: {
           optimizer: {
             enabled: true,
             details: {
               yul: !process.env.CI,
             },
           },
+          evmVersion: "paris",
```
chechu, line 160: `paris` didn't include `PUSH0`

```suggestion
          evmVersion: "shanghai",
```

GitGuru7: I think Shanghai EVM version `PUSH0` opcode is not supported by every chain.

kkirka: I didn't do my own research, but AFAIK all chains we use already support `PUSH0`. I may be wrong, though. @GitGuru7 @0xlucian do you guys know any particular chain from our list that doesn't support `PUSH0`?

kkirka: Looks like opBNB mainnet and testnet do not support shanghai yet. They plan to add support for it in a mid-June release. I'd merge the `paris` version for now, unless we can specify evm version per chain (I haven't found the info on how to do that though – looks like compilation step is chain-agnostic in hardhat).

chechu (approved):

chechu (approved):

PR #398, opened 2024-05-02
Add archived pool lens versions

coreyar:
## Description

Adds PoolLens archive for previous version of the PoolLens so that it can be used to query past block.


chechu (approved):

chechu (commented):
We are using `legacy` instead of `archive` in `governance-contracts` [repo](https://github.com/VenusProtocol/governance-contracts). Are they different concepts?

coreyar:
> We are using `legacy` instead of `archive` in `governance-contracts` [repo](https://github.com/VenusProtocol/governance-contracts). Are they different concepts?

they are the same I can change to legacy

chechu (approved):

PR #417, opened 2024-06-26
[VEN-2663] Two Kinks Interest Rate Model

web3rover:
## Description

Resolves #

## Checklist


contracts/TwoKinksInterestRateModel.sol
```diff
@@ -0,0 +1,210 @@
...
+ * @author Venus
+ * @notice An interest rate model with two different steep increase each after a certain utilization threshold called **kink** is reached.
+ */
+contract TwoKinksInterestRateModel is InterestRateModel, TimeManagerV8 {
+    ////////////////////// SLOPE 1 //////////////////////
+
+    /**
+     * @notice The multiplier of utilization rate per block or second that gives the slope 1 of the interest rate
+     */
+    int256 public immutable MULTIPLIER_PER_BLOCK_OR_SECOND;
```
Debugger022, line 19: I will say to not make these variables immutable and make a setter also for them otherwise we will have to make new deployment every time we want to change a single value.

Debugger022: I do not remember it but got your point.

Debugger022: Also if we deploy new IR then we do not need to check in each VToken that the same IR is not in use by any other VToken to prevent other VTokens to update IR too. It will be more error free and time saving.

GitGuru7 (dismissed):

chechu (approved):

PR #419, opened 2024-07-04
Increase max collateral factor

coreyar:
## Description

Increase Max Collateral Factor to 95%


kkirka (approved):

contracts/ComptrollerStorage.sol
```diff
@@ -109,7 +109,7 @@ contract ComptrollerStorage {
     uint256 internal constant MAX_CLOSE_FACTOR_MANTISSA = 0.9e18; // 0.9
 
     // No collateralFactorMantissa may exceed this value
-    uint256 internal constant MAX_COLLATERAL_FACTOR_MANTISSA = 0.9e18; // 0.9
+    uint256 internal constant MAX_COLLATERAL_FACTOR_MANTISSA = 0.95e18; // 0.95
```
GitGuru7, line 113: Should we keep it as it is, or revert it?

coreyar: We will keep it

PR #442, opened 2024-09-24
Add forked deployment

coreyar:
## Description

Add supports for doing a "dry run" of deployments using hardhat. Require setting URI at variables `ARCHIVE_NODE_<NETWORK>` and setting the network with `HARDHAT_FORK_NETWORK`.

Without `HARDHAT_FORK_NETWORK` set it will run like normal. With `HARDHAT_FORK_NETWORK` set it runs on hardhat but overrides hardhat values with those of the forked networks


deploy/012-transfer-pools-ownership.ts
```diff
@@ -40,7 +17,7 @@ const transfer2StepOwnerships = async (contractNames: string[], targetOwner: str
     const pendingOwner = await contract.pendingOwner();
 
     let tx;
-    if (owner !== targetOwner && pendingOwner !== targetOwner) {
+    if (owner.toLowerCase() !== targetOwner.toLowerCase() && pendingOwner.toLowerCase() !== targetOwner.toLowerCase()) {
```
kkirka, line 20: I believe the convention should be to use checksummed addresses everywhere. This is what ethers returns in the calls, so we could avoid case conversions in this case.

kkirka (approved):
LGTM

deploy/012-transfer-pools-ownership.ts
coreyar: Updated

kkirka (dismissed):

hardhat.config.ts
```diff
@@ -21,44 +21,70 @@ import { convertToUnit } from "./helpers/utils";
...
           "node_modules/@venusprotocol/oracle/deployments/arbitrumsepolia",
           "node_modules/@venusprotocol/protocol-reserve/deployments/arbitrumsepolia",
+          "node_modules/@venusprotocol/governance-contracts/deployments/arbitrumsepolia",
+        ],
+        arbitrumone: [
+          "node_modules/@venusprotocol/protocol-reserve/deployments/arbitrumone",
+          "node_modules/@venusprotocol/governance-contracts/deployments/arbitrumsepolia",
         ],
-        arbitrumone: ["node_modules/@venusprotocol/protocol-reserve/deployments/arbitrumone"],
         basesepolia: [
```
chechu, line 88: Don't we need to add the governance deployments to base and base sepolia?

hardhat.config.ts
```diff
@@ -69,6 +95,15 @@ extendConfig((config: HardhatConfig) => {
         ],
       },
     };
+    if (process.env.HARDHAT_FORK_NETWORK) {
+      config.external.deployments!.hardhat = [
```
chechu, line 99: Would this alternative work?

`config.external.deployments!.hardhat = config.external.deployments[process.env.HARDHAT_FORK_NETWORK]`

coreyar: If a network is not being forked, we don't want to use external deployments. For example testing

chechu (approved):

PR #453, opened 2024-11-13
Transient storage implementation

Debugger022:
## Description

Resolves #

## Checklist


contracts/Gateway/ReentrancyGuardStorage.sol
chechu, line 1: would it make sense to move this to the solidity-utilities package, with the rest of the new files related to the feature that can be reused?

PR #497, opened 2025-02-25
[VEN-2887] - Vaults(4626) implementation for vTokens

Debugger022:
## Description

This PR introduces the implementation of the ERC-4626 wrapper for Venus markets for isolated pools, enabling seamless integration with external DeFi protocols that follow the ERC-4626 standard. This enhancement includes the following changes:

ERC-4626 Implementation: Developed an ERC-4626 compliant wrapper for Venus VToken markets, standardizing deposits, withdrawals, and balance calculations.

Reward Distribution: Integrated reward claiming from RewardsDistributor, transferring rewards to the designated recipient and handling updates for Protocol Share Reserve (PSR) contracts.

Factory Contract: Introduced a factory contract to facilitate the deployment of ERC-4626 wrappers for different Venus markets.

MockVenusERC4626 Contract: Added a mock implementation of the ERC-4626 wrapper for testing purposes, enabling controlled simulations of deposits, withdrawals, and reward interactions.

Test Cases: Added comprehensive unit tests covering deposits, withdrawals, reward claiming, and recipient scenarios (EOA vs. PSR).

Hardhat Config File : fixed the external deployment dependency of arbitrumone for governance-contracts


contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,260 @@
...
+        uint256 assetsBalance = convertToAssets(balanceOf(receiver));
+        return cash < assetsBalance ? cash : assetsBalance;
+    }
+
+    /// @notice Returns the maximum amount of shares that can be redeemed.
+    /// @dev Redemption is limited by the available cash in the vault.
+    /// @param receiver The address of the account redeeming.
+    /// @return The maximum number of shares that can be redeemed.
+    function maxRedeem(address receiver) public view override returns (uint256) {
+        uint256 cash = VTOKEN.getCash();
```
chechu, line 214: Shouldn't we check if the REDEEM action is paused in the vToken (similar to the check we are performing in the deposit flow)?

contracts/ERC4626/Base/ERC4626Factory.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.13;
```
chechu, line 2: 0.8.25?

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,135 @@
...
+
+    /// @notice The reward distributor contract
+    RewardDistributorInterface public immutable REWARD_DISTRIBUTOR;
+
+    /// -----------------------------------------------------------------------
+    /// Storage variables
+    /// -----------------------------------------------------------------------
+
+    /// @notice Maps underlying asset to the corresponding vToken
+    mapping(ERC20 => VToken) public underlyingToVToken;
```
chechu, line 39: This mapping is already stored in the `poolRegistry` associated with the comptroller. We could call something like `comptroller.poolRegistry().getVTokenForAsset(comptroller, asset)`, to get the vToken associated with the asset in that comptroller

This way, we could remove this mapping and the function `updateUnderlyingToVToken`

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,135 @@
...
+/// @notice Factory for creating VenusERC4626 contracts
+contract VenusERC4626Factory is ERC4626Factory {
+    using SafeERC20 for ERC20;
+
+    /// -----------------------------------------------------------------------
+    /// Immutable params
+    /// -----------------------------------------------------------------------
+
+    /// @notice The XVS token contract
+    ERC20 public immutable XVS;
```
chechu, line 23: I think we don't need `XVS` and `REWARD_DISTRIBUTOR` as attributes in this factory

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,260 @@
...
+    ///      If there is any XVS balance after claiming, it is transferred to the predefined reward recipient.
+    ///      Emits a `ClaimRewards` event with the amount of XVS claimed.
+    function claimRewards() external {
+        REWARD_DISTRIBUTOR.claimRewardToken(address(this));
+        uint256 xvsBalance = XVS.balanceOf(address(this));
+
+        if (xvsBalance > 0) {
+            XVS.safeTransfer(REWARD_RECIPIENT, xvsBalance);
+        }
+        emit ClaimRewards(xvsBalance);
```
chechu, line 112: I think we should send them to the ProtocolShareReserve, with a new Income type: `ERC4626_WRAPPER_REWARDS`. I wouldn't hardcode the PSR address, ofc, we can keep the `REWARD_RECIPIENT` attribute, but we could "try-catch" call `REWARD_RECIPIENT.updateAssetsState` at the end of this function

Debugger022: Okay, I am checking for the rewardRecipient contract size if it exist then I am calling the `REWARD_RECIPIENT.updateAssetsState`
[3306ff2](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/3306ff24862b3f0f8cb2a3544d6054571336d16e)

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,260 @@
...
+
+    /// @notice Checks if a specific vault action is paused.
+    /// @dev Reverts if the given vault action is paused.
+    /// @param action The vault action to check (e.g., deposit, mint, withdraw, redeem).
+    function _checkVaultActionPaused(VaultAction action) private view {
+        if (_vaultPaused[action]) {
+            revert VaultActionPaused(action);
+        }
+    }
+}
```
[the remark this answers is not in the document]
Debugger022: [3306ff2](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/3306ff24862b3f0f8cb2a3544d6054571336d16e)

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @param /*account*/ The address of the account.
+    /// @return The maximum amount of assets that can be deposited.
+    function maxDeposit(address /*account*/) public view virtual override returns (uint256) {
+        if (COMPTROLLER.actionPaused(address(VTOKEN), Action.MINT)) {
+            return 0;
+        }
+
+        uint256 supplyCap = COMPTROLLER.supplyCaps(address(VTOKEN));
+        uint256 totalSupply_ = VTOKEN.totalSupply();
+        return supplyCap > totalSupply_ ? supplyCap - totalSupply_ : 0;
```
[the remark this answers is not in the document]
Debugger022: [3306ff2](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/3306ff24862b3f0f8cb2a3544d6054571336d16e)

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @dev Calls the `claimRewardToken` function on the reward distributor contract to claim rewards.
+    ///      If there is any XVS balance after claiming, it is transferred to the predefined reward recipient.
+    ///      Emits a `ClaimRewards` event with the amount of XVS claimed.
+    function claimRewards() external {
+        REWARD_DISTRIBUTOR.claimRewardToken(address(this));
+        uint256 xvsBalance = XVS.balanceOf(address(this));
+
+        if (xvsBalance > 0) {
+            XVS.safeTransfer(REWARD_RECIPIENT, xvsBalance);
+        }
```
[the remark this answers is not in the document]
Debugger022: Added functionality to support multiple rewardDistributors, [3306ff2](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/3306ff24862b3f0f8cb2a3544d6054571336d16e)

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -72,108 +56,106 @@ contract VenusERC4626 is ERC4626, Ownable {
...
 
         uint256 assets = previewRedeem(shares);
-        beforeWithdraw(assets, shares);
+        beforeWithdraw(assets);
         _withdraw(_msgSender(), receiver, redeemer, assets, shares);
         return assets;
     }
 
+    function totalAssets() public view virtual override returns (uint256) {
+        return (VTOKEN.balanceOf(address(this)) * VTOKEN.exchangeRateStored()) / 1e18;
```
chechu, line 156: What if we used `EXP_SCALE` from `lib/constants.sol` instead of hardcoding 1e18 in this file?

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,244 @@
...
+
+                // Try to update the asset state on the recipient if reward recipient is a protocol share reserve
+                if (REWARD_RECIPIENT.code.length != 0) {
+                    IProtocolShareReserve(REWARD_RECIPIENT).updateAssetsState(
+                        address(COMPTROLLER),
+                        address(rewardToken),
+                        IProtocolShareReserve.IncomeType.ERC4626_WRAPPER_REWARDS
+                    );
+                }
+            }
```
GitGuru7, line 102: Would be better to do a low level static call to check if the function exists?

```
(bool success, ) = target.staticcall(abi.encodeWithSelector(selector));
return success;
```

This way the function does not have to revert if the reward recepient is a contract but not the PSR.

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    /// -----------------------------------------------------------------------
+    /// External functions
+    /// -----------------------------------------------------------------------
+
+    /// @inheritdoc ERC4626Factory
+    function createERC4626(ERC20 asset) external virtual override returns (ERC4626 vault) {
+        VToken vToken = VToken(
+            PoolRegistryInterface(COMPTROLLER.poolRegistry()).getVTokenForAsset(address(COMPTROLLER), address(asset))
+        );
```
GitGuru7, line 51: Maybe we can optimise the constructor args of venusErc4626 contract to accept address of vToken to avoid importing the whole vToken contract here.

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,244 @@
...
+
+    /// @notice Thrown when a redemption exceeds the maximum redeemable shares.
+    /// @dev This error is triggered if the redemption amount is greater than `maxRedeem(owner)`.
+    error ERC4626__RedeemMoreThanMax();
+
+    /// @notice Constructor to initialize the Venus ERC4626 vault.
+    /// @param vToken_ The Venus vToken contract.
+    /// @param rewardRecipient_ The address receiving the rewards.
+    constructor(
+        VToken vToken_,
```
GitGuru7, line 63: maybe convert this to address to avoid importing the complete vToken contract in the factory.

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -0,0 +1,244 @@
...
+        COMPTROLLER = IComptroller(address(vToken_.comptroller()));
+        REWARD_RECIPIENT = rewardRecipient_;
+    }
+
+    /// @notice Claims rewards from all reward distributors associated with the VToken and transfers them to the reward recipient.
+    /// @dev Iterates through all reward distributors fetched from the comptroller, claims rewards, and transfers them if available.
+    function claimRewards() external {
+        RewardsDistributor[] memory rewardDistributors = COMPTROLLER.getRewardDistributors();
+
+        for (uint256 i = 0; i < rewardDistributors.length; i++) {
```
GitGuru7, line 82: Would make sense to analyse the gas cost here to ensure we never run out of gas for 3-4 reward distributors. This could become a problem if there are more added in the future.

package.json
```diff
@@ -51,7 +51,8 @@
     "@venusprotocol/solidity-utilities": "2.0.3",
     "ethers": "^5.7.0",
     "hardhat-deploy": "^0.11.14",
-    "module-alias": "^2.2.2"
+    "module-alias": "^2.2.2",
+    "solmate": "^6.8.0"
```
GitGuru7, line 55: Are we using solmate?

contracts/ERC4626/VenusERC4626.sol
```diff
@@ -1,55 +1,39 @@
...
 import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { ERC4626 } from "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
+import { ensureNonzeroAddress } from ".././lib/validators.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { VToken } from "../VToken.sol";
-import { IComptroller, Action } from "./Interfaces/IComptroller.sol"; // Venus Comptroller interface.
-import { RewardDistributorInterface } from "./Interfaces/IRewardDistributor.sol"; // Interface for claiming rewards.
-import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { IComptroller, Action } from "./Interfaces/IComptroller.sol";
+import { RewardsDistributor } from ".././Rewards/RewardsDistributor.sol";
```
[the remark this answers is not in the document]
Debugger022: we are getting distributors through `COMPTROLLER.getRewardDistributors()` and it returns array of distributors. so we had to use `RewardsDistributor instead of `IRewardsDistributor`.

kkirka (commented):
1. My main concern wrt the proposed implementation is upgradeability. VenusERC4626 uses quite complex logic involving reward claims, supply/borrow caps, PSR integration, etc. Although we try to preserve interface compatibility between the protocol revisions, there is no guarantee we won't at some point introduce additional functionality that needs to be taken into account in this wrapper (e.g. per-user caps, defferent reward mechanics, more restrictions on mints and redemptions, etc.). A beacon proxy per ERC4626 could be a solution here (we could still use the create2 approach), but AFAIU we won't be able to use immutables in this case.

2. AFAIU, the proposal is to deploy a factory per pool. So it's neither a "one single factory for everything" nor a "separate contract per each vToken". I wonder why we opt for something in between here, as it seems we're taking the worst from both worlds – we  still need to deploy and maintain multiple contracts on chain _and_ we need to deal with create2 and its limitations. Maybe there's some flexibility  we want to achieve this way (like setting different reward receivers for different pools)?

chechu:
> 2. AFAIU, the proposal is to deploy a factory per pool. 

I think we could improve this with these changes:

* replace the `COMPTROLLER` attribute with a `POOL_REGISTRY` attribute (there is a pool registry per chain, so we would have a VenusERC4626 factory per chain)
* accept the VToken address in the `createERC4626` function of the factory. We would require that the VToken is valid in the pool registry

Debugger022:
> 1. My main concern wrt the proposed implementation is upgradeability. VenusERC4626 uses quite complex logic involving reward claims, supply/borrow caps, PSR integration, etc. Although we try to preserve interface compatibility between the protocol revisions, there is no guarantee we won't at some point introduce additional functionality that needs to be taken into account in this wrapper (e.g. per-user caps, defferent reward mechanics, more restrictions on mints and redemptions, etc.). A beacon proxy per ERC4626 could be a solution here (we could still use the create2 approach), but AFAIU we won't be able to use immutables in this case.
> 2. AFAIU, the proposal is to deploy a factory per pool. So it's neither a "one single factory for everything" nor a "separate contract per each vToken". I wonder why we opt for something in between here, as it seems we're taking the worst from both worlds – we  still need to deploy and maintain multiple contracts on chain _and_ we need to deal with create2 and its limitations. Maybe there's some flexibility  we want to achieve this way (like setting different reward receivers for different pools)?

I have made ERC4626 contracts beacon upgradable and Factory contract is also upgradable now.
[091b01f](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/091b01f2cd6d9994c6a6519c140b7219af5e5c32)
[8e15e08](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/8e15e085f2d764a2e628fd99786a3edc01d8d3fa)

Debugger022:
> > 2. AFAIU, the proposal is to deploy a factory per pool.
> 
> I think we could improve this with these changes:
> 
> * replace the `COMPTROLLER` attribute with a `POOL_REGISTRY` attribute (there is a pool registry per chain, so we would have a VenusERC4626 factory per chain)
> * accept the VToken address in the `createERC4626` function of the factory. We would require that the VToken is valid in the pool registry

1. Refactored.
2. I am using comptroller and asset address  in the `createERC4626` function of the factory.

package.json
Debugger022: Nope, removed, [2a4290e](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/2a4290ee5f4307d10cad51f0c0e2faf6fda31346)

GitGuru7 (dismissed):

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        address vToken = poolRegistry.getVTokenForAsset(comptroller, asset);
+        if (vToken == address(0)) {
+            revert VenusERC4626Factory__VTokenNonexistent();
+        }
+
+        vault = ERC4626Upgradeable(
+            address(
+                new BeaconProxy(
```
kkirka, line 103: I thought the plan was to use a factory and create2, do you think using `new BeaconProxy{salt: salt}(...)` here can cause any issues?

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,112 @@
...
+
+        emit RewardRecipientUpdated(rewardRecipient, newRecipient);
+        rewardRecipient = newRecipient;
+    }
+
+    /// @notice Creates a VenusERC4626 vault for a given asset and comptroller
+    /// @param comptroller The comptroller's address to which asset belongs
+    /// @param asset The ERC20 asset address belongs to the comptroller
+    /// @return vault The deployed VenusERC4626 vault
+    function createERC4626(address comptroller, address asset) external returns (ERC4626Upgradeable vault) {
```
chechu, line 91: Wouldn't it be more intuitive if we receive the address of a VToken here. This factory will deploy VenusERC4626 contracts, and each VenusERC4626 contract is associated with a VToken

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,112 @@
...
+
+import { AccessControlledV8 } from "@venusprotocol/governance-contracts/contracts/Governance/AccessControlledV8.sol";
+import { PoolRegistryInterface } from ".././Pool/PoolRegistryInterface.sol";
+import { VenusERC4626 } from "./VenusERC4626.sol";
+import { ensureNonzeroAddress } from ".././lib/validators.sol";
+
+/// @title VenusERC4626Factory
+/// @notice Factory for creating VenusERC4626 contracts
+contract VenusERC4626Factory is Ownable2StepUpgradeable, AccessControlledV8 {
+    using SafeERC20Upgradeable for ERC20Upgradeable;
```
chechu, line 19: Do we need this? I don't see any use of `ERC20Upgradeable` in this contract

contracts/ERC4626/VenusERC4626Factory.sol
Debugger022: Initially, I was thinking to use the factory and create2 for the 4626 vaults but to make vaults upgradable, I chose to opt for the beacon approach in the same way we are doing it with vTokens and comptrollers.
I guess it would not cause any issue, please let me know if you have any thoughts.

contracts/ERC4626/Interfaces/IProtocolShareReserve.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.25;
+
+interface IProtocolShareReserve {
+    /// @notice it represents the type of vToken income
+    enum IncomeType {
+        SPREAD,
+        LIQUIDATION,
+        ERC4626_WRAPPER_REWARDS
```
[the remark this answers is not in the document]
Debugger022: Yes, I will raise an PR for the same in the reserve-protocol repo.

contracts/ERC4626/VenusERC4626Factory.sol
Debugger022: Removed, [df78de4](https://github.com/VenusProtocol/isolated-pools/pull/497/commits/df78de4e335ede29fe3a680af369626be05aa1e7)

contracts/ERC4626/VenusERC4626Factory.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    /// @notice Creates a VenusERC4626 vault for a given asset and comptroller
+    /// @param vToken The vToken address to create the vault
+    /// @return vault The deployed VenusERC4626 vault
+    function createERC4626(address vToken) external returns (ERC4626Upgradeable vault) {
+        ensureNonzeroAddress(vToken);
+
+        VTokenInterface vToken_ = VTokenInterface(vToken);
+
+        if (poolRegistry.getPoolByComptroller(address(vToken_.comptroller())).comptroller == address(0)) {
```
chechu, line 89: I would check this:

```
vToken == poolRegistry.getVTokenForAsset(vToken_.comptroller(), vToken_.underlying());
```

To be sure the provided vToken is included in a valid pool registered in our registry. With the current check, any contract returning a valid comptroller in the `comptroller()` function would pass the check.

contracts/ERC4626/Interfaces/IComptroller.sol
chechu, line 1: We could simply define the functions we'll use:

```
function actionPaused(address market, Action action) external view returns (bool);
function getRewardDistributors() external view returns (RewardsDistributor[] memory);
function supplyCaps(address) external view returns (uint256);
```

I wouldn't redefine the Action enum. We could import and use it from `contracts/ComptrollerInterface.sol`. Less code should be better :) wdyt?

contracts/ERC4626/VenusERC4626Factory.sol
Debugger022: Used `getVTokenForAsset` instead of `getPoolByComptroller` to verify.

chechu:
This code was moved to its own repository: https://github.com/VenusProtocol/ERC4626
The initial commit of that repository contains the codebase of this pull request: https://github.com/VenusProtocol/ERC4626/commit/771f392e011841a60fee68abe31cbf225b628b91

Debugger022:
closing it: https://github.com/VenusProtocol/isolated-pools/pull/497#issuecomment-3056665824

PR #501, opened 2025-04-07
[VEN-3163]: Add CheckpointRateModel

kkirka:
This pull request introduces the CheckpointRateModel contract, a new interest rate model for the Venus Protocol. The CheckpointRateModel dynamically switches between two interest rate models based on a predefined checkpoint timestamp. This feature is particularly useful for adjusting interest rates in response to scheduled changes in network conditions, such as block time updates.


contracts/InterestRateModels/CheckpointRateModel.sol
```diff
@@ -0,0 +1,73 @@
...
+    function getSupplyRate(
+        uint256 cash,
+        uint256 borrows,
+        uint256 reserves,
+        uint256 reserveFactorMantissa,
+        uint256 badDebt
+    ) external view override returns (uint256) {
+        return _getCurrentRateModel().getSupplyRate(cash, borrows, reserves, reserveFactorMantissa, badDebt);
+    }
+
```
chechu, line 65: I would also implement `utilizationRate`, used by the UI, AFAIK.

contracts/InterestRateModels/CheckpointRateModel.sol
```diff
@@ -0,0 +1,73 @@
...
+import { InterestRateModel } from "./InterestRateModel.sol";
+
+/**
+ * @title Venus CheckpointRateModel Contract
+ * @notice A contract that applies a rate model based on whether a certain checkpoint
+ *   in time has passed. Useful to adjust interest rates for block time changes in
+ *   the underlying networks, if the block time update is scheduled in advance.
+ * @author Venus
+ */
+contract CheckpointRateModel is InterestRateModel {
```
chechu, line 14: `JumpRateModelV2` inherits `TimeManagerV8` too. Should `CheckpointRateModel` implement the public views, to offer the same interface?

contracts/InterestRateModels/CheckpointRateModel.sol
kkirka: Should we put `utilizationRate` to `InterestRateModel` interface in this case? So that all rate models implement it

contracts/InterestRateModels/CheckpointRateModel.sol
kkirka: I don't think so. It's a different rate model, so it's ok for it to not implement the same interface as `JumpRateModelV2`, as long as it implements `InterestRateModel`. We can have a checkpoint rate model migrating from JumpRateModel to TwoKinksRateModel at a certain timestamp, there's no restriction that the two rate model kinds must be the same.

kkirka:
I don't have a strong opinion regarding reimplementing each InterestRateModel function explicitly vs proxying the calldata in the fallback. The latter is a bit less explorer-friendly but this way consumers can work with `CheckpointRateModel` as if it was the original rate model, without the checkpoint.

contracts/InterestRateModels/CheckpointRateModel.sol
chechu: Theoretically, there could be an interest rate model not based on the utilization rate. But if that would simplify the integration with third parties (i.e. the dapp), it lgtm

kkirka: I added a separate interface (0e67188ec7968ea854159997e35e41db112b981a) and the support for utilizatoin rate (9b9a5d814b5631b4e2ccc82a00b75565638b0bfb). Let me know if you think this is the right direction. I'll also raise an alternative PR based on fallback – it seems a bit more universal and it is less likely to break any clients.

kkirka:
Closing in favor of https://github.com/VenusProtocol/venus-protocol/pull/576

PR #505, opened 2025-04-15
refactor: stop using BSC_BLOCKS_PER_YEAR in tests

kkirka:
Problem: Some of the tests rely on BSC_BLOCKS_PER_YEAR to be a constant value, e.g. rewards and interests in tests are computed under this assumption. Adjusting test cases for a different number of blocks per year every time this value is changed is an unnecessary overkill, since the logic itself doesn't depend on a particular value assumed by the test suite.

Solution: Introduce a separate constant, DEFAULT_BLOCKS_PER_YEAR, that should be used in tests going forward.

### DIFF
--- a/helpers/deploymentConfig.ts
+++ b/helpers/deploymentConfig.ts
@@ -147,6 +147,7 @@ export enum InterestRateModels {
 
 const ANY_CONTRACT = ethers.constants.AddressZero;
 
+export const DEFAULT_BLOCKS_PER_YEAR = 10_512_000; // assuming a block is mined every 3 seconds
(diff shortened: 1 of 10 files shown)


helpers/deploymentConfig.ts
```diff
@@ -147,6 +147,7 @@ export enum InterestRateModels {
 
 const ANY_CONTRACT = ethers.constants.AddressZero;
 
+export const DEFAULT_BLOCKS_PER_YEAR = 10_512_000; // assuming a block is mined every 3 seconds
```
chechu, line 150: Let's set `21_024_000`, one block every 1.5 seconds, to check the effects of the next upgrade on BNB Chain

PR #508, opened 2025-04-15
[VEN-3163]: Add checkpoint rate models

kkirka:
This PR adds deployments for [CheckpointView contracts](https://github.com/VenusProtocol/venus-protocol/pull/576) pointing to different rate models – ones with the old block rate before the specified timestamp, and ones with the new block rate after the specified timestamp.

* Old rate models were deployed previously, https://github.com/VenusProtocol/isolated-pools/pull/507 added the old blocks per year suffixes. Note that the "old rate model" is not necessarily equal to the currently configured rate model since duplicates were removed. However, they should be identical in terms of the parameters.
* New rate models (with the reduced block time) were deployed in https://github.com/VenusProtocol/isolated-pools/pull/503

This PR contains two deployment scripts – one that deploys the CheckpointView contracts and __generates a setter contract,__ and one that deploys this setter contract (for each network). These setter contracts will be used in the VIP to configure the new rate models.


deploy/020-rate-model-checkpoint.ts
```diff
@@ -0,0 +1,108 @@
...
+import { writeGeneratedContract } from "../helpers/writeFile";
+
+const checkpoints = {
+  bsctestnet: {
+    at: 1744097580, // 2025-04-08 07:33:00 AM UTC,
+    fromBlocksPerYear: 10_512_000,
+    toBlocksPerYear: 21_024_000,
+  },
+  bscmainnet: {
+    at: 1745895600, // 2025-04-29 03:00:00 AM UTC
```
kkirka, line 15: Should be 05:05:00 GMT

kkirka, line 15: Should be

```suggestion
    at: 1745903100, // 2025-04-29 05:05:00 AM UTC
```

chechu (approved):

GitGuru7 (approved):

deploy/020-rate-model-checkpoint.ts
kkirka: Fixed here (https://github.com/VenusProtocol/isolated-pools/pull/508/commits/79de9e7236bd7bbacb333f3981fcbe1c0af48083) and redeployed

PR #559, opened 2026-08-18
Feat/hub spoke pools

Debugger022:
## What

Adds `SpokeComptroller`, a fork of `Comptroller` for hub-funded spoke pools, behind its own beacon. Two deploy scripts: the comptroller, and a second `PoolRegistry` instance this pool registers in instead of the isolated-pools one. No existing pool is affected.

`SpokePoolLens` is a new lens for the spoke-only state. No shared contract is modified.

Nothing is wired or listed by this PR. The oracle, `addPool`, the ACM grants and the markets all belong to the listing VIP.

## Decisions

- **Forked instead of extending `Comptroller`.** `Comptroller` is shared by every pool in this repo, on this chain and others. Spoke-only policy does not belong in it, and a shared-beacon upgrade would put that policy on all live pools at once. The fork has its own beacon, so blast radius is zero.
- **The interface is inherited, not copied.** `SpokeComptrollerInterface is ComptrollerInterface`, so both comptrollers share one definition of the function surface and it cannot drift. Only the events and errors are declared in the spoke interface. Separate interface for `SpokeComptroller` also works 
- **Prime integration removed.** Not used by this pool, and it freed ~1,067 bytes of the size budget.
- **The seven `*Verify` hooks kept as empty no-ops.** `VToken` calls all seven. Removing them would force a separate `VToken` implementation for this pool, so keeping them lets the pool reuse the shared `VToken` beacon.
- **Revert strings replaced with custom errors.** Smaller bytecode, and consistent with the newer contracts in the repo.
- **Bounded pricing applies only to the collateral-factor paths.** Borrow and redeem value collateral at the low end of the asset's recent price window and debt at the high end, so a deviating price can only reduce borrowing capacity. The liquidation-threshold paths stay on spot, because they decide liquidation routing and bounding them would distort eligibility.
- **Optimizer `runs: 30` for this contract instead of moving code into an external library.** With bounded pricing added, the contract no longer fits under EIP-170 at the default 200 runs. Measured: `runs: 30` costs +1,901 gas on a borrow (under 1.5%), while a linked library costs at least 2,600 gas for the cold `DELEGATECALL` alone, so it was both more expensive and a larger architectural change. The library stays the fallback if headroom ever runs out, since it frees kilobytes rather than bytes.
- **Storage layout unchanged in shape.** The reserved gap was shrunk by the number of new slots, so the fork occupies the same slots as `ComptrollerStorage`.
- **A `PoolRegistry` instance of its own, not a row in the isolated-pools registry.** That registry is the directory every consumer iterates to answer which pools exist, so a pool whose supply, borrow and liquidation sides are all restricted to known accounts would need a special case in each of them. A second instance also keeps the two products independently upgradeable and permissioned. The chain's existing `DefaultProxyAdmin` is reused, so there is no second admin to track.
- **The pool is deliberately absent from `poolConfig`.** The standard scripts iterate that list and would deploy this pool on the shared beacon under the same artifact name; this script would then skip it as already deployed, leaving a pool that looks correct but runs the wrong implementation.

## New behaviour

- Per-market supply allowlist, disabled by default
- Pool-wide liquidation allowlist, disabled by default
- Per-market liquidation incentive, used to route an unhealthy account between `liquidateAccount` and `healAccount`
- Deviation-bounded pricing on the borrow capacity paths, with a setter for the oracle address
- `enterMarketBehalf(address,address)`, so a router supplying through `mintBehalf` enters the supplier rather than itself. ACM-gated, one market per call to stay under EIP-170. `enterMarkets` is untouched, so self entry stays permissionless
- `liquidateAccount` refreshes interest and prices before its routing snapshot, as `healAccount` already did, so the two do not decide on different views of the same position
- `SpokePoolLens` reports the spoke-only state: bounded oracle and liquidation allowlist per pool, liquidation incentive and supply allowlist per market. The rest of its surface mirrors `PoolLens` under the same names and struct shapes, so reading a spoke pool takes one address rather than two. Probed by `staticcall`, so a comptroller without them reports absence instead of reverting the read

## Fuzz tests

Foundry fuzz and invariant tests for the new spoke behaviour, added in [f888bb3](https://github.com/VenusProtocol/isolated-pools/commit/f888bb332698813d472faba621000f042fa858ea) under `tests/foundry/Spoke/`. Foundry reads only `tests/foundry/`, so the Hardhat setup is unchanged. All pass. Not in CI yet.

Run: `forge test --match-path "tests/foundry/Spoke/*"`

**Fuzz tests** (random inputs, 256 runs each)

- Pool liquidation incentive is accepted only at 1.05x or more
- Market incentive and protocol seize share always respect each other's limits
- After a 15% to 45% price drop, a liquidator receives at least the value it repaid
- Bounded borrowing power is never above borrowing power at spot price
- A successful borrow never exceeds the collateral factor limit at spot price

**Invariants** (random sequences of supply, borrow, repay, redeem, transfer, all three liquidation paths, reward claims, price moves and allowlist or incentive changes across 8 accounts, checked after every action, 256 runs x 100 actions)

- No mint credits an account off the supply allowlist
- No liquidation succeeds for a liquidator off the liquidation allowlist
- `liquidateAccount` and `healAccount` never both accept the same account
- A liquidator never receives less than the debt it repaid
- No successful action leaves an account in shortfall
- `SpokePoolLens` pending rewards equal what a claim pays
- Bounded borrowing power is never above spot
- vToken total supply equals the sum of holder balances
- Cash a market counts is really held by the market
- Rewards paid never exceed what was emitted
- `SpokePoolLens` data matches the pool and markets

## Known Gaps — Deferred

None of the items below are permissionless. Every one sits behind `onlyOwner` or the
`AccessControlManager`, so they are governance and configuration guardrails rather than direct
vulnerabilities. They are recorded here so the listing VIP and any future change to this contract
start from a known baseline.

### 1. DBO and Comptroller oracles are not cross-checked

`DeviationBoundedOracle.RESILIENT_ORACLE` is immutable, while `SpokeComptroller.oracle` is settable
through `setPriceOracle`. Nothing verifies the two refer to the same `ResilientOracle`.

If they ever diverge, borrowing power is priced through the bounded oracle while liquidation routing
and seize math are priced through the comptroller's own oracle, with no revert and no event to make
it visible.

**Action:** the listing VIP verifies the pairing at deployment. A future change could enforce it in
`setPriceOracle` and `setDeviationBoundedOracle`, at a cost in contract size (see item 7).

### 2. No maximum limit on liquidation incentives

Both `setLiquidationIncentive` and `setMarketLiquidationIncentive` enforce only a floor. Neither
rejects a value that is too high.

This is a risk-parameter decision and is intentionally left to governance, matching how the shared
`Comptroller` treats it. One consequence is specific to this fork and worth being aware of when
setting the value: the per-market incentive divides into `maxClearableDebt`, so an unusually high
incentive on one collateral market shrinks that figure for every borrower holding it, which shifts
those accounts from `liquidateAccount` toward `healAccount` and therefore toward bad debt.

**Action:** none in code. Set per-market incentives deliberately.

### 3. Unset DBO oracle causes an unnamed revert

Borrow and redeem fail closed while `deviationBoundedOracle` is unset, because
`_updateProtectionStates` calls into the zero address. The revert carries no named error, so the
failure is correct but opaque.

The window in which this is reachable is between proxy initialization and the listing VIP, and the
VIP sets the oracle before the pool serves any action.

**Action:** none. Failing closed is the intended behaviour; only the error message is unhelpful, and
only in a window nobody transacts in.

### 4. An empty liquidation allowlist can freeze liquidations

Enabling `isLiquidationAllowlistEnabled` with no approved liquidators blocks every seizure in the
pool. This also covers `healAccount`, so the keeper relied on to record bad debt has to be
allowlisted too, not just the liquidators.

**Action:** the listing VIP adds every liquidator and bad-debt keeper before enabling the allowlist,
in the same transaction.

### 5. `PoolData.liquidationIncentive` still reports the pool-wide value

`liquidationIncentiveMantissa()` resolves against `msg.sender`, so a lens reading it receives the
pool-wide value, and that is what `PoolData.liquidationIncentive` continues to mean. Changing the
meaning of an existing field would break current consumers.

**Action:** none. The per-market value is now reported as
`VTokenMetadata.liquidationIncentiveMantissa`, so a consumer reading a spoke pool takes that field.
The on-chain seize amount was never affected: `liquidateCalculateSeizeTokens` and `VToken._seize`
both resolve the collateral market's own incentive.

### 6. `maxLoopsLimit` needs re-validation

Borrow, redeem and transfer now walk an account's markets twice per call, once for
`_updatePrices` and once for `_updateProtectionStates`, on top of the reward-distributor loops. The
deployed limit of 100 was chosen before the second walk existed, and neither oracle loop is bounded
by `_ensureMaxLoops`.

**Action:** measure the worst-case borrow against a realistic market count and adjust the deployed
value if the gas ceiling, rather than the configured limit, is what actually binds.

### 7. Contract size

`SpokeComptroller` compiles to **24,551 bytes against the 24,576-byte EIP-170 limit, leaving 25
bytes**, at the `runs: 30` override this contract already needs in order to fit. Any further change
has to be measured before it is written.

Moving the liquidity snapshot into an external library would free kilobytes rather than bytes, but it
has been left out for now to keep `SpokeComptroller` aligned with the IL `Comptroller` it was forked
from.

One related note: all four compiler blocks disable the Yul optimizer under `CI`. Compiled that way
the contract measures 25,657 bytes, over the limit, while the production build is under it. Any size
check that runs in CI is therefore measuring a binary that is never deployed.

## Listing vip important notes

1. In the listing VIP, PoolRegistry.addMarket must set vTokenReceiver to the Spoke YieldGroup (not the Timelock), so the initial supply is minted in the YieldGroup's name and the Hub stays the market's only supplier.
2. Six new ACM role strings, verbatim (`enterMarketBehalf`, `setMarketLiquidationIncentive`, `setSupplyAllowlistEnabled`, `setAllowedSupplier`, `setLiquidationAllowlistEnabled`, `setAllowedLiquidator`), plus the inherited `setActionsPaused(address[],uint256[],bool)` trap where the registered string deliberately disagrees with the real ABI (`uint8[]`). Deriving that one from the ABI gives you a grant that never matches.
3. Pool-wide incentive ≥ `1.05e18`, now enforced, so this one fails loudly.
4. DBO: three checks, two of which fail silently. A market with no `setTokenConfig` runs with zero protection and no revert; the caching-flag pairing; the `RESILIENT_ORACLE` pairing.
5. Supply allowlist after `addMarket`, with the YieldGroup granted in the same VIP.
6. Liquidation allowlist: liquidators and the bad-debt keeper, since it gates `healAccount` too.
7. Per-market incentives: the floor, the `0`-is-a-sentinel rule, and the ordering trap where lowering the pool-wide value after a market's seize share was raised is never re-checked.
8. `SpokePoolRegistry`: `acceptOwnership`, since the deploy script only nominates, plus ACM grants in both directions. Existing wildcard grants name the isolated-pools registry's address, so none carry over and `addPool` reverts without new ones. The exact list is in the deploy script's closing comment.
9. ProtocolShareReserve holds a single `poolRegistry` address. Pointing it at `SpokePoolRegistry` stops the existing isolated pools from taking income; not pointing it there stops this pool. The multi-registry change ships from another repo and has to be live first. Pinned in `tests/hardhat/Fork/HubSpoke/psrRegistryConflict.ts`.
10. `enterMarketBehalf`: grant the role only to a router that passes its own caller as `account`. Anything else can enter markets for anyone.

11. Wildcard ACM grants are live on `SpokeComptroller` the moment it is deployed, before the listing VIP. The role is `keccak256(address(0), functionSig)`, names no target, and the fork reuses the upstream signatures verbatim, so every inherited setter except `setForcedLiquidation` is covered. Held today on BSC mainnet ACM by NormalTimelock, FastTrackTimelock, the Guardian multisig and the isolated-pools `PoolRegistry`. Re-check the holders before listing.


contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+        _checkActionPauseState(vToken, Action.MINT);
+
+        if (!markets[vToken].isListed) {
+            revert MarketNotListed(address(vToken));
+        }
+
+        // `minter` is the account credited with the newly minted vTokens, not necessarily the account paying for
+        // them: `mintBehalf` lets a third party fund a mint attributed to someone else. Metering the recipient is
+        // what bounds the market's supply, so the payer is deliberately not checked.
+        if (isSupplyAllowlistEnabled[vToken] && !isAllowedSupplier[vToken][minter]) {
```
fred-venus, line 336: ok, to make sure i understand correctly, its on purpose to only check in `preMintHook`, i mean if some whitelisted addr mint some share and then transfer it to someone else, the current logic gonna allow it.

fred-venus: ofc by design, only borrow-out asset needed to be whitelist (fund solely from LH):
1. only yieldGroup contract gonna be whitelisted
2. yieldGroup won't send the token to anyone else

Debugger022: Yes, that's right, and the mint gate is the only one. `isAllowedSupplier` gates exactly one thing, `preMintHook`. Transfers and seizures both move vTokens without consulting it.
For the liquidity side that's fine, and it holds structurally rather than by convention:

- the adapter never transfers vTokens. Its only transfer is the underlying in `withdraw`, so they leave the YieldGroup by burn only.
- `YieldGroup.sweep` refuses a registered resource, and `removeResource` requires the receipt balance to already be zero, so neither can move a live position.
- the group has no borrow path at all, so there is nothing to seize.

contracts/Spoke/SpokeComptroller.sol
GitGuru7, line 1: IMO, it would be better to avoid creating another copy of the Comptroller. Instead, I think we should consider making the changes in the main Comptroller itself, so we don’t end up maintaining two copies and adding more technical debt.

I’m thinking about two possible solutions:

**1. Add the whitelisting logic to the main Comptroller** (Preferred)

The Comptroller is already quite large, so yes adding more logic directly wont be possible. However, this could also be a good opportunity to move the setter functions into facets exposed through a simple fallback + facet selector mechanism. This should give us enough additional size headroom, while also leaving more room for future changes.

We would likely need to do this eventually anyway, and IMO it’s not a particularly large change to include in the audit scope, especially considering code change scope.

**2. Split the main Comptroller into a base Comptroller**

If we don’t want to add the whitelisting logic directly to the main Comptroller, another option would be to split the current Comptroller into a base Comptroller, making the relevant functions virtual. We could then have two Comptroller implementations on top of it, one for the spoke and one for the existing implementation.

In both scenarios, I think the code would be easier to maintain, while the audit scope would remain more or less the same.

And in either case, we should also move the other changes currently included in the spoke Comptroller, Replacing the require with the custom revert and adding the per-market risk factors.

Debugger022: I raised this early during the PRD discussion: https://app.notion.com/p/PRD-Hub-Funded-Spoke-pools-381b4e5771bb81a09edfe5a253f47174?d=3bab4e5771bb8058b4c1001ca9c8c9cf&source=copy_link#14edc7706ccd4001ac9afc1790d8b4b2

There are two main reasons why I decided to keep the Spoke Comptroller as a fork:

* **Avoid affecting live Core pools:** `ComptrollerBeacon` is shared across other chains (Ethereum, Base, etc.). Adding Spoke specific logic to the main Comptroller would require upgrading all of them, even though this new policy is only needed for Spoke pools.

* **Keep the implementations aligned:** `SpokeComptrollerInterface` inherits from `ComptrollerInterface`, and we have tried to keep the Spoke and main Comptroller logic as aligned as possible. This should also make it easier to have a common audit approach for both implementations in the future.

**Contract size is another constraint:** The Spoke additions only fit because Prime is dropped and the contract compiles at runs: 30. Neither of these constraints is acceptable for a shared implementation used by all Core markets.

image

To keep the audit and review scope focused, I created the fork in the first [commit](https://github.com/VenusProtocol/isolated-pools/pull/559/commits/de23dbc868b66f50e591fbec3941681abeb2825c), so the actual review can start from the next commit. If this approach causes any issues, we can instead use the fork as the base branch and raise the PR on top of it.

With all these factors in mind, if the team still prefers a different approach, I’m open to making the necessary changes.

cc: @fred-venus , @zed-venus

contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+            }
+        }
+
+        if (!market.accountMembership[borrower]) {
+            revert MarketNotCollateral(vTokenCollateral, borrower);
+        }
+
+        // Every seizure reaches this hook with the account that receives the collateral, whichever entry point it
+        // came from, so this is the check that actually enforces the allowlist.
+        _checkLiquidationAllowed(liquidator);
```
GitGuru7: can we rename to _checkLiquidatorAllowlisted ?
_checkLiquidationAllowed sounds very generic in general

contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+     * @custom:error NonzeroBorrowBalanceAfterLiquidation is thrown if the orders do not clear every borrow
+     * @custom:error SnapshotError is thrown if some vToken fails to return the account's supply and borrows
+     * @custom:error PriceError is thrown if the oracle returns an incorrect price for some asset
+     * @custom:access Not restricted while the liquidation allowlist is disabled, otherwise restricted to the accounts
+     *   on it
+     */
+    function liquidateAccount(address borrower, LiquidationOrder[] calldata orders) external {
+        // Every order seizes, so `preSeizeHook` would reject a caller that is not allowed anyway. Checking at the
+        // entry keeps the two batch operations symmetric and fails before any interest is accrued or debt repaid.
+        _checkLiquidationAllowed(msg.sender);
```
GitGuru7: Since this isn’t a common entry point that users would go through, I don’t think we really need the early check. The risk is already handled at the point where it applies, which should be sufficient.

contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+     *  spot on both legs otherwise, including for an asset it holds no configuration for. A deviating print can
+     *  therefore only ever shrink an account's borrowing capacity, never inflate it. Under the liquidation threshold
+     *  both legs are spot, because those snapshots route an unhealthy account between `liquidateAccount` and
+     *  `healAccount` and set how much of its debt healing repays, which has to track the live price.
+     * @param asset Address for asset to query prices for
+     * @param weighting Which risk parameter weights the position being valued
+     * @return collateralPrice Price valuing the collateral held in the market
+     * @return debtPrice Price valuing the debt owed to the market
+     */
+    function _safeGetPrices(
```
GitGuru7: can we use boundedPrice or weightedUnderlyingPrice ?

contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+        return (Exp({ mantissa: collateralPriceMantissa }), Exp({ mantissa: debtPriceMantissa }));
+    }
+
+    /**
+     * @dev Returns the risk parameter that weights a market's collateral in a liquidity snapshot
+     * @param asset Address for asset whose parameter to read
+     * @param weighting Which of the two parameters to read
+     * @return The market's collateral factor or liquidation threshold, as an exponential
+     */
+    function _weight(VToken asset, WeightFunction weighting) internal view returns (Exp memory) {
```
GitGuru7: weightingFactor ?

contracts/Spoke/SpokeComptroller.sol
```diff
@@ -0,0 +1,1921 @@
...
+     *  frame exceeds the stack the legacy code generator can address, and this repo does not enable `viaIR`.
+     *  `snapshot` is a memory reference, so the accumulated fields are updated in place.
+     * @param snapshot The snapshot to accumulate into
+     * @param asset The market to value
+     * @param account The account whose position in the market is being valued
+     * @param weighting Which per-market risk parameter weights the collateral
+     * @return weightedVTokenPrice Value of one vToken after the risk weight, which prices a hypothetical redeem
+     * @return debtPrice Price valuing the market's debt, which prices a hypothetical borrow
+     */
+    function _accumulateMarket(
```
GitGuru7: _accumulateMarketPosition ?

fred-venus:
i dont see anything uncommon, but do have 2 points raising from ai, i checked make sense and seems inherited from existing codebase  

1. Pre-accrual liquidation snapshot

> liquidateAccount determines eligibility before accruing the relevant markets, while execution uses the updated state and skips the liquidity check. In theory it could happen that the ahead checking is eligible to liquidate but back to healthy after.

2. Fallback liquidation incentive

> Lowering the pool-wide liquidation incentive can violate LI >= 1 + protocolSeizeShare for markets using the fallback value, the new fallback mechanism should enforce this invariant when updating the pool-wide LI

contracts/Spoke/SpokeComptroller.sol
zed-venus: Let's go with the fork for this release.

Three reasons, all scope and risk rather than style:

- Blast radius. ComptrollerBeacon is shared across chains. Putting spoke-only policy in the main Comptroller means upgrading every live Core pool for a feature only spoke pools use. Not a risk I want to carry for this launch.
- Audit scope. A facet or base-contract refactor of the shared Comptroller is its own project with its own review. Bundling it here widens the audit surface.
- Reversibility. A fork behind its own beacon is easy to walk back. A shared-implementation change is not.

fred-venus:
Continuing the topic in group, we will need to deploy a new spoke pool registry and will need to update the addr in protocol share reserve as well. However, without any change to current PSR impl this will revert the existing isolated pool when sending out the revenue because of this check

CleanShot 2026-09-02 at 13 55 05@2x


So we might need to update the logic of PSR a bit to unblock the existing isolated pool, tho its sunset at product level but lets at least make sure there is no unexpected revert onchain

contracts/Spoke/SpokeComptroller.sol
Debugger022: I mentioned reason above in the comment 

> // Every order seizes, so `preSeizeHook` would reject a caller that is not allowed anyway. Checking at the
> // entry keeps the two batch operations symmetric and fails before any interest is accrued or debt repaid.

but removing this it would free few bytes

contracts/Spoke/SpokeComptroller.sol
Debugger022: going with `_safeGetUnderlyingPrices`

contracts/Spoke/SpokeComptroller.sol
Debugger022: since it's a getter so `_getWeightingFactor`

contracts/Spoke/SpokeComptroller.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/2300cc5300d0725fad7cd6403a3ca4ff38ff8f6b

contracts/Spoke/SpokeComptroller.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/ce9aa070dea46d023bfeec52ea2cd04bbe8ab28f

contracts/Spoke/SpokeComptroller.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/ce9aa070dea46d023bfeec52ea2cd04bbe8ab28f

contracts/Spoke/SpokeComptroller.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/ce9aa070dea46d023bfeec52ea2cd04bbe8ab28f

contracts/Spoke/SpokeComptroller.sol
Debugger022: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/ce9aa070dea46d023bfeec52ea2cd04bbe8ab28f

Debugger022:
> i dont see anything uncommon, but do have 2 points raising from ai, i checked make sense and seems inherited from existing codebase
> 
> 1. Pre-accrual liquidation snapshot
> 
> > liquidateAccount determines eligibility before accruing the relevant markets, while execution uses the updated state and skips the liquidity check. In theory it could happen that the ahead checking is eligible to liquidate but back to healthy after.
> 
> 2. Fallback liquidation incentive
> 
> > Lowering the pool-wide liquidation incentive can violate LI >= 1 + protocolSeizeShare for markets using the fallback value, the new fallback mechanism should enforce this invariant when updating the pool-wide LI

@fred-venus both points are correct and yes both are inherited from existing codebase

### 1. Pre-accrual liquidation snapshot

`healAccount` accrues interest and refreshes prices before it takes its snapshot. `liquidateAccount` takes the same kind of snapshot on stored balances and the last recorded price, and since the orders then run with `skipLiquidityCheck`, nothing re-tests eligibility on live state afterwards.

**Fix:** the accrue + `updatePrice` loop is pulled out of `healAccount` into a shared `_refreshMarkets`, and `liquidateAccount` now calls it before its snapshot. No contract size issue, Size actually drops, since the duplicated loop and a second `getAssetsIn` walk are gone.
fixed: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/395e6a64bfb99ff2eecc53e3ce7545937ba06697

### 2. Fallback liquidation incentive

The gap is real, but I would keep it out of the contract:

- It is already documented in the NatSpec on `setLiquidationIncentive`: a market whose seize share is raised above the default needs an incentive of its own.
- The other side of the invariant is enforced. `VToken.setProtocolSeizeShare` checks `1e18 + newShare` against the calling market's *effective* incentive, so raising a share cannot break it. Only lowering the pool-wide value afterwards can, and only for a market that has no incentive of its own.
- Markets are expected to be configured with their own liquidation incentive, so the fallback is not what they read in practice.
- For any market that does fall back, this is covered at the VIP level: a VIP lowering the pool-wide incentive checks it against `1e18 + protocolSeizeShareMantissa()` of every fallback-using market.
- Enforcing it on chain means looping `allMarkets` in the setter, which would cause contract size issue

Worth noting the consequence there is liveness, not loss: if the invariant is violated the liquidator receives less than it repaid, so liquidations of that market stop rather than anything being drained.

Debugger022:
> Continuing the topic in group, we will need to deploy a new spoke pool registry and will need to update the addr in protocol share reserve as well. However, without any change to current PSR impl this will revert the existing isolated pool when sending out the revenue because of this check
> 
> CleanShot 2026-09-02 at 13 55 05@2x
> So we might need to update the logic of PSR a bit to unblock the existing isolated pool, tho its sunset at product level but lets at least make sure there is no unexpected revert onchain

- PSR updated logic PR:https://github.com/VenusProtocol/protocol-reserve/pull/168
- Deploy script for pool registry: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/89cb3b741aa8e2ac5da6b0df1404ad034fe7e56b

fred-venus:
One more thing 

CleanShot 2026-09-02 at 19 32 32@2x


We probably want to poolLen as well to reflect DBO and per asset liquidation incentive

Debugger022:
> One more thing
> 
> CleanShot 2026-09-02 at 19 32 32@2x
> We probably want to poolLen as well to reflect DBO and per asset liquidation incentive

also this 
image

Fixed: https://github.com/VenusProtocol/isolated-pools/pull/559/commits/b261a7b17f7589db587f82eec0ffb2a15dc919cc

fred-venus (dismissed):

contracts/Lens/PoolLens.sol
```diff
@@ -368,8 +378,18 @@ contract PoolLens is ExponentialNoError, TimeManagerV8 {
             vTokens: vTokenMetadataItems,
             priceOracle: address(comptrollerViewInstance.oracle()),
             closeFactor: comptrollerViewInstance.closeFactorMantissa(),
+            // The pool-wide value on either kind of pool: a spoke resolves this getter against `msg.sender`, and a
+            // lens is not one of its markets. Per market, read `VTokenMetadata.liquidationIncentiveMantissa`.
             liquidationIncentive: comptrollerViewInstance.liquidationIncentiveMantissa(),
-            minLiquidatableCollateral: comptrollerViewInstance.minLiquidatableCollateral()
+            minLiquidatableCollateral: comptrollerViewInstance.minLiquidatableCollateral(),
+            deviationBoundedOracle: _probeAddress(
```
fred-venus: is this to decide whether its isolated pools or spoke pools ?

i am thinking why dont we just redeploy a new PoolLens with clean implementation, right now we already have new poolRegistry, its natural to think we should have a poolLens as well.


So basically 2 sets of contracts in parallel so they dont confuse ppl

fred-venus: also for VTokenBeacon lets deploy a new one, so that in the future if ever we do the change its not gonna impact existing isolated pools

Debugger022: The lens isn't bound to a registry, it takes `poolRegistryAddress` as a call argument, and it's a plain non-proxy contract we version by redeploying (R1/R2 archived, this PR archives R3), so the separation already exists at the address level. But fair point on the confusion, I'll add a separate SpokePoolLens and leave PoolLens for isolated. That also drops the staticcall probing.

Debugger022: Agreed, will deploy a separate VToken beacon for the spoke pool.

Debugger022: - [feat(lens): add a dedicated lens for spoke pools](https://github.com/VenusProtocol/isolated-pools/pull/559/commits/3058dbfbedc1007ed4b441f9d6776023ab972c05)

- [feat(spoke): add a deploy script for the spoke vToken beacon](https://github.com/VenusProtocol/isolated-pools/pull/559/commits/6989643f9805bea28398e27483a8e904eaaa5a54)

Debugger022:
I noticed unlistMarket doesn’t check for leftover supply. we could add:  
`if (VToken(market).totalSupply() != 0) revert MarketStillHasSupply();`
This would prevent any remaining collateral from being locked forever. However, governance would check totalSupply before unlisting. This behaviour is inherited from isolated pools. Should we add this check as a safeguard?

Also, unlistMarket only sets isListed = false, it doesn’t remove the market from allMarkets. So if we ever want to add the same market again, addMarket will revert with MarketAlreadyListed because the address still exists in allMarkets. 

LMK your thoughts
cc: @fred-venus @zed-venus

fred-venus:
> I noticed unlistMarket doesn’t check for leftover supply. we could add:  `if (VToken(market).totalSupply() != 0) revert MarketStillHasSupply();` This would prevent any remaining collateral from being locked forever. However, governance would check totalSupply before unlisting. This behaviour is inherited from isolated pools. Should we add this check as a safeguard?
> 
> Also, unlistMarket only sets isListed = false, it doesn’t remove the market from allMarkets. So if we ever want to add the same market again, addMarket will revert with MarketAlreadyListed because the address still exists in allMarkets.
> 
> LMK your thoughts cc: @fred-venus @zed-venus

i think its fine lets not add it

1. based on the experience we never really unlist market, usually just pause actions and then gradually sunset, in the past only market like luna is literally "unlisted"
2. if wanna unlist cant really count on zero the supply, there will always be token

fred-venus:
can probabaly fix the CI

